PR #3, opened 2017-09-15
Targetting netstandard2.0

### LINKED ISSUE #2: MissingMethodException when running on .NET Core 2.0
The usage of `EntityTagHeaderValue` is like this:
https://github.com/SixLabors/ImageSharp.Web/blob/9edd7eb054348503f6bb87fa13e03bc8ecfb87ae/src/ImageSharp.Web/Middleware/ImageContext.cs#L215

When an application using ImageSharp.Web is using a newer version of `Microsoft.Net.Http.Headers`, literally 2.0.0, then this constructor is not available anymore as it has been converted to a `StringSegment`.
It results in a `MissingMethodException` whenever the middleware is used.

There are two solutions:
- Target 2.0.0 only and reference the new packages
- Multi-target 1.4 and 2.0.0 with a conditional statement to switch to the other implementation

Please advise on your preferred solution and I will implement it. At this point I personally don't think supporting 1.4 still makes sense as you released after 2.0.

sebastienros:
Preventing MissingMethodException on EntityTagHeaderValue.ctor(string)

Fixes #2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,8 @@ branches:
 
 script:
   - dotnet restore
-  - dotnet test tests/ImageSharp.Web.Tests/ImageSharp.Web.Tests.csproj -c Release
+  - dotnet test tests/ImageSharp.Web.Tests/ImageSharp.Web.Tests.csproj -c Release -f "netcoreapp2.0"
+  - dotnet test tests/ImageSharp.Web.Tests/ImageSharp.Web.Tests.csproj -c Release -f "netcoreapp1.1"
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,3 +40,9 @@ addons:
     build_command_prepend: "dotnet restore"
     build_command:   "dotnet build -c Release"
     branch_pattern: coverity_scan
+  apt:
+    sources:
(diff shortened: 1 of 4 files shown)


sebastienros:
I assume it will be available on myget first. When it's there I will validate.

tocsoft:
@sebastienros thanks for this 👍 
Your right once we merge in the fix it will be available on our myget feed.

@JimBobSquarePants I'll leave this one up to you to make a decision on as it changes our target version.

sebastienros:
I published the appveyor package artifact to my own myget feed and could confirm it working fine now.

JimBobSquarePants:
How have we suddenly got a missing method? This was all working.

JimBobSquarePants:
Ok, I see now. I'd rather support both if I can until we absolutely cannot.

sebastienros:
Sure ... updating the PR then. I assume you want to run the test on both too ...

JimBobSquarePants:
Yes please. Thanks!

JimBobSquarePants:
Looks like we're missing the actual code change for the constructor no?

sebastienros:
You don't need a code change, this is why it was only found at runtime. Because on netcore2 it's requiring a StingSegment and when built the string is converted automatically.

JimBobSquarePants:
Ah! I see, sorry early morning here. Yeah, I guess they thought, major version and all that.

sebastienros:
@JimBobSquarePants ready

PR #4, opened 2017-09-15
Refactor options registration

sebastienros:
Option configuration services shouldn't be singletons as they should be discarded once called.

Using TryAdd as calling AddImageSharpCore several times (common pattern) would otherwise overwrite custom options.

### DIFF
--- a/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Options;
 using SixLabors.ImageSharp.Web.Caching;
 using SixLabors.ImageSharp.Web.Commands;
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
I don't follow. Why would you call the method several times?

sebastienros:
In a modular application (see Orchard) multiple module might call into `AddImageSharpCore` to set different options. One might set the `MaxCacheDays` property, and another could set something else that is available in options. That's a common pattern in ASP.NET Core, and that how options are designed.

Yes there is a single option instance, but we register services that configure this object, and we can register several of these services. If we register yours, the default one, every time, then it will revert the changes that the previous one did. So by doing `TryAdd` we ensure that you are not overriding any option setup that was already defined, or that if called another time it won't erase the new values.

Also we use Transient as there is no need to keep the service alive in the DI. Again the service is transient, the option is unique and kept alive. You register an initializer here, not an option type.

JimBobSquarePants:
Ah right... That all makes sense to me then. I'll download this tonight just to double check and merge it then. 

Thanks again!

tocsoft:
@sebastienros sorry to have kept you waiting, all merged now.

We will try and arrange a release of a second beta soon, we'll probably time it to release with some fixes to ImageSharp to reduce package churn.

Issue #5, opened 2017-09-21
ImageSharp.Web 


tocsoft:
### Description
This issue is designed to capture the architecture and tasks required to build an extensible library for parsing URL API instructions and caching the result.

Feel free to discuss below. We'll assign specific issues for functional components as we determine them.

### Basic Requirements 

- [x] Asp.NET Core Middleware to capture valid requests
- [x] Extensible configuration for Caching, ImageServices, and Processing. Must be secure.
- [x] `IImageService` implementations for physical file system and remote images. 
- [x] `IDistributedCache` implementation for physical file system that can be switched out for Redis, Azure, etc. How do we handle cleanup, Response codes?
- [x] Fast, low collision hashing of URL for cache implementations.

### Pipeline

The following gives an overall summary of the image processing pipeline.

1. Request is intercepted by Middleware (Authorisation?)
2. `IImageService` is assigned using URL prefix pattern matching (Regex?)
3. `IImageService` determines validity of request and attempts to return stream containing image.
4. `IDistributedCache` implementation checks if file exists. (Redirect if exists?)
5. Not cached? URL querystring params are parsed for matching processor.
6. Image is processed by matches via fluent API. 
7. Result is cached and image served.
8. Cleanup of expired cached files. (How?, When?)

# Edit

**Updated Requirements and Pipeline below**

### Basic Requirements 

- [x] Asp.NET Core Middleware to capture valid requests
- [x] Extensible configuration for Parsing, Resolving, Processing, and Caching. Must be secure
- [ ] <strike>Configuration for handling EXIF metadata</strike> Managed via `OnBeforeSave`
- [x] `IRequestParser` implementation for parsing requests for command string that can be switched out
- [x] `IImageResolver` implementation collection for physical file system that 
- [x] Extensible `CommandParser` implementation for converting commands into real values for processing. 
- [x] `IImageWebProcessor` for Resize
- [ ] <strike>`IImageWebProcessor` for Quality</strike> Managed by configuring individual formats.
- [x] `IImageWebProcessor` for Format
- [x] `IImageWebProcessor` for BackgroundColor
- [x] `IImageCache` implementation for physical file system that can be switched out for Azure, etc.
- [x] Fast, low collision hashing of URL for cache implementations
- [x] `ILogger` based logging
- [ ] <strike/>`TagHelper` for generating `img`, `picture` markup</strike> Out of scope.

### Pipeline

The following gives an overall summary of the image processing pipeline.

[...]

PR #6, opened 2017-10-09
Add OnBeforeSave for metadata. Touch #5

JimBobSquarePants:
### Prerequisites

### Description
This PR adds a new configurable action to the processing pipeline `OnBeforeSave` which passes the current `FormattedImage` instance for final adjustment before encoding to the output stream. This allows manipulation of EXIF metadata or anything else you can think of that cannot be handled via the URL API.


src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -174,6 +174,7 @@ public class ImageSharpMiddleware
                 using (var image = FormattedImage.Load(this.options.Configuration, inBuffer))
                 {
                     image.Process(this.logger, this.processors, commands);
+                    this.options.OnBeforeSave(image);
```
tocsoft: probably want to call this using this pattern `this.options.OnBeforeSave?.Invoke(image);` in the event `OnBeforeSave` is ever manually set to null.

tocsoft: We probably want to make sure we do that for all the extension points form the `options`

JimBobSquarePants: I don't like the idea of swallowing up exceptions like that. If someone manually sets it to null it should throw. 

I'm actually following the pattern from the MS staticfile middleware.

tocsoft: Its not an exception really as you wouldn't be swallowing anything its just skipping running the delegate if its but unset.

JimBobSquarePants: Here's what I mean. 

https://github.com/aspnet/StaticFiles/blob/826e89a2d1b325485819040ff41b9948afc91e06/src/Microsoft.AspNetCore.StaticFiles/StaticFileContext.cs#L285

Can't find a unit test to copy either. We do need to do some others though.

JimBobSquarePants:
@tocsoft Finally got some decent integration tests running. Fixes some bugs in the process.

PR #7, opened 2017-10-15
Lock identical requests to reduce processing overheads

JimBobSquarePants:
### Prerequisites

### Description
This PR adds new classes `AsyncKeyLock`, `DoormanPool`, and `Doorman` that allow the locking of processing requests based on the url. 

The `AsyncKeyLock` will allow multiple requests using different keys to process simultaneously ensuring throughput is high but will ensure that duplicate requests are blocked until the original request has processed.

By doing this we reduce processing overheads and remove the chance of file locks when saving the cached file.

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -27,7 +27,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
         /// <summary>
         /// The keylock used for limiting identical requests
         /// </summary>
-        private readonly AsyncKeyLock keyLock = new AsyncKeyLock();
+        private readonly IAsyncKeyLock asyncKeyLock;
 
         /// <summary>
         /// The function processing the Http request.
@@ -79,6 +79,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
         /// <param name="resolvers">A collection of <see cref="IImageResolver"/> instances used to resolve images</param>
         /// <param name="processors">A collection of <see cref="IImageWebProcessor"/> instances used to process images</param>
         /// <param name="cache">An <see cref="IImageCache"/> instance used for caching images</param>
+        /// <param name="asyncKeyLock">An <see cref="IAsyncKeyLock"/> instance used for providing locking during processing</param>
         public ImageSharpMiddleware(
             RequestDelegate next,
             IOptions<ImageSharpMiddlewareOptions> options,
@@ -86,7 +87,8 @@ namespace SixLabors.ImageSharp.Web.Middleware
             IUriParser uriParser,
             IEnumerable<IImageResolver> resolvers,
             IEnumerable<IImageWebProcessor> processors,
-            IImageCache cache)
+            IImageCache cache,
+            IAsyncKeyLock asyncKeyLock)
         {
             Guard.NotNull(next, nameof(next));
             Guard.NotNull(options, nameof(options));
@@ -95,6 +97,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
             Guard.NotNull(resolvers, nameof(resolvers));
             Guard.NotNull(processors, nameof(processors));
             Guard.NotNull(cache, nameof(cache));
+            Guard.NotNull(asyncKeyLock, nameof(asyncKeyLock));
 
             this.next = next;
             this.options = options.Value;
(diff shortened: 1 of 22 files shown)


tocsoft:
Do we want to wrap this into an injectable interface? then it would be possible to add a cross process/azure scaled out locking option for when the cache is in a shared location between instances.

```c#
interface IImageProcessingLocker{
Task<IDisposable> Lock(string key);
}```

**Comment by JimBobSquarePants** [comment]
Aye maybe... If you were to build your own version how would you keep allocations down?

In my head we could maybe use a named Semaphore using the key to make the lock cross-process but then we wouldn't be able to pool the used instances. 
Maybe we should stick to the interface like you say and let someone else worry about trying to create that.... I find multithreading hard.

**Comment by JimBobSquarePants** [comment]
@tocsoft Actually it looks like named semaphores are not implemented on Unix so maybe we're out of luck there and can't actually do this.

https://github.com/dotnet/corefx/pull/2796

**Comment by tocsoft** [comment]
its not just cross process that needs worrying about its cross machine... there are lots of options to allow this sort of blocking... be it writing a file to disk and monitoring it or using a shared redis server to orchestrate the locks. 

But that's a decision for other developers to fight with if/when they ever want to implement the process locker. I say we should not worry about allocations other people might introduce by swapping out our implementation and just wrap the `AsyncKeyLock` in an interface that a user can implement if desired.

**Comment by JimBobSquarePants** [comment]
That sounds like a good plan to me 👍 I've added the configurable interface and moved the classes to the Caching namespace.

File: src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs [hunk 1]
```diff
@@ -127,81 +135,85 @@ public class ImageSharpMiddleware
...
-            try
-            {
-                inBuffer = await resolver.ResolveImageAsync(context, this.logger);
-                if (inBuffer == null || inBuffer.Length == 0)
+                if (resolver == null || !await resolver.IsValidRequestAsync(context, this.logger))
                 {
-                    // Log the error but let the pipeline handle the 404
-                    this.logger.LogImageResolveFailed(imageContext.GetDisplayUrl());
+                    // Nothing to do. call the next delegate/middleware in the pipeline
                     await this.next(context);
```
tocsoft: not sure we should be calling `this.next()` inside the lock... I'm pretty sure that will cause all requests to become single threaded... we want to move the `await this.next(context);` to outside the lock with a flag to determine weather to call it instead.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -127,81 +135,85 @@ public class ImageSharpMiddleware
...
+                byte[] outBuffer = null;
+                MemoryStream outStream = null;
+                try
+                {
+                    inBuffer = await resolver.ResolveImageAsync(context, this.logger);
+                    if (inBuffer == null || inBuffer.Length == 0)
+                    {
+                        // Log the error but let the pipeline handle the 404
+                        this.logger.LogImageResolveFailed(imageContext.GetDisplayUrl());
+                        await this.next(context);
```
tocsoft: as above should be moved outside the lock... might be easier to move this whole method down into a new internal method that returns true if request was handled else false... then we can call `this.next(context)` based on the outcome instead of inline inside the lock like we have here.

JimBobSquarePants: Done 👍

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -123,86 +131,110 @@ public class ImageSharpMiddleware
...
+                            outBuffer = BufferDataPool.Rent(outLength);
+                            await outStream.ReadAsync(outBuffer, 0, outLength);
+
+                            DateTimeOffset cachedDate = await this.cache.SetAsync(key, outBuffer, outLength);
+                            await this.SendResponse(imageContext, key, cachedDate, outBuffer, outLength);
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        this.logger.LogImageProcessingFailed(imageContext.GetDisplayUrl(), ex);
```
tocsoft: I feel we're missing a `processRequest = false;` in here

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -123,86 +131,110 @@ public class ImageSharpMiddleware
...
-                // Buffer should have been rented in IImageResolver
-                BufferDataPool.Return(inBuffer);
-                BufferDataPool.Return(outBuffer);
-            }
+        /// <summary>
+        /// Calls the request delegate to transfer the request to the next middleware in the pipeline.
+        /// </summary>
+        /// <param name="context">The current HTTP request context</param>
+        /// <returns>The <see cref="Task"/></returns>
+        private async Task Next(HttpContext context)
```
tocsoft: how's this better than calling the  `await this.next(context);` directly?

PR #16, opened 2018-02-11
Improve/fix FormatHelpers.GetExtension URI parsing

jhdrn:
### Prerequisites

### Description

This is related to https://github.com/OrchardCMS/OrchardCore/issues/1472

I've made some changes to the FormatHelpers.GetExtension method to make it "safer":
- At first it checks the query string for the "format" parameter, then it searches the path of the URI for a known file extension.
- It only matches an extension preceeded by a dot.
- It now correctly selects the most matching extension in case of multiple extensions with the same character sequence, for example "bm" and "bmp". Previously it would match "bm" even if the URI contained "bmp".

### DIFF
--- a/src/ImageSharp.Web/Helpers/FormatHelpers.cs
+++ b/src/ImageSharp.Web/Helpers/FormatHelpers.cs
@@ -64,11 +64,6 @@ namespace SixLabors.ImageSharp.Web.Helpers
                 }
             }
 
-            if (extension != null && !path.EndsWith(extension))
-            {
-                return null;
-            }
-
             return extension;
         }
 
--- a/src/ImageSharp.Web/Helpers/FormatHelpers.cs
+++ b/src/ImageSharp.Web/Helpers/FormatHelpers.cs
@@ -4,6 +4,8 @@
 using System;
 using System.IO;
 using System.Linq;
+using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Primitives;
 using SixLabors.ImageSharp;
 using SixLabors.ImageSharp.Formats;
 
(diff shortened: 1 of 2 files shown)


src/ImageSharp.Web/Helpers/FormatHelpers.cs
```diff
@@ -34,14 +36,25 @@ public static string GetContentType(Configuration configuration, string key)
         /// <returns>The <see cref="string"/></returns>
         public static string GetExtension(Configuration configuration, string uri)
         {
+            string[] parts = uri.Split('?');
+            if (parts.Length > 1)
+            {
+                StringValues ext;
+                if (QueryHelpers.ParseQuery(parts[1]).TryGetValue("format", out ext))
```
JimBobSquarePants: Should use `FormatWebProcessor.Format` rather than the inline string. We can inline that out param also to keep it pretty.

jhdrn: Done! :)

jhdrn:
Great! Do you have any plans for a new release with these changes?

JimBobSquarePants:
There's an automatic build in our MyGet repo you can consume for now. We'll release a new beta though once some dependency changes have been completed. Hopefully within the next month.

jhdrn:
Ok, sounds great, but unfortunately the MyGet repo (https://www.myget.org/gallery/imagesharp) does not seem to be public.  I'm kind of in need of this fix so the question is whether you plan on making the MyGet repo public or if I should resort to make an own package for now?

JimBobSquarePants:
Och the link in the readme is wrong!

It's...

https://www.myget.org/feed/sixlabors/package/nuget/SixLabors.ImageSharp.Web

I'll get it updated.

jhdrn:
Ah, that explains it :) 

The last package in that feed is 1.0.0-ci0032 published on Fri, 17 Nov 2017 21:11:30 GMT though, so it doesn't look like the automatic build did trigger.

JimBobSquarePants:
Cripes, everything has gone wrong. There's a bug in the tool we're consuming to produce version numbers.

1.0.0-ci0010 is actually the latest 🙁

tocsoft:
The build numbers are now fixed. You will be wanting any build >= 1.0.0-dev000064

PR #21, opened 2018-03-31
Fix DI and make BufferPool injectable

JimBobSquarePants:
### Prerequisites

### Description
This PR does three things.

1. Makes `BufferDataPool` injectable via `IBufferDataPool` so we can provide configurable pooling behaviour.
2. Fixes the factory overloads for injecting the various singletons as they were providing the wrong method siganture (concrete vs interface).
3. Renames `IUriParser` to `IRequestParser` so that that we can allow parsing requests by means other than URI commands without confusion.


src/ImageSharp.Web/Memory/BufferDataPool.cs
```diff
@@ -6,9 +6,9 @@
 namespace SixLabors.ImageSharp.Web.Memory
 {
     /// <summary>
-    /// Provides a resource pool that enables reusing instances of value type arrays for image data <see cref="T:Byte[]"/>.
+    /// Provides a resource pool that enables reusing arrays for transporting encoded image data.
     /// </summary>
-    public class BufferDataPool
+    public class BufferDataPool : IBufferDataPool
```
tocsoft: shouldn't this be replaced with `MemoryManager` from `ImageSharp` and flow the specified manager back into the configuration used during loading/resizing etc? 

This would require the  `Allocate`/`IBuffer<T>` instances being made public.

tocsoft: However I wouldn't bother blocking the PR for this change, just something we aught to be aiming towards.

JimBobSquarePants: Should’ve ran the tests before pushing. Poor form on my part. MemoryManager didn’t exist when I first wrote the web part but you’re right. I should refactor to now use it. Should be easy enough.

PR #30, opened 2018-08-06
WIP API Updates

JimBobSquarePants:
### Prerequisites

### Description
Updates to allow new beta.

- [x] Remove `ILogger` from `IImageResolver` and cleanup unused code.
- [x] Update dependencies and StyleCop
- [x] Replace `IBufferManager` to use new `SixLabors.Memory` APIs
- [x] Replace `IImageResolver` with `IImageProvider` 
- [x] New type `IImageResolver`. #32 
- [x] Use pool-backed output stream #33


JimBobSquarePants:
@antonfirsov I'd love a hand implementing the new `SixLabors.Memory` APIs if you have time.

antonfirsov:
@JimBobSquarePants I hope I'll have some time for this tomorrow. Want to update docs and samples first.

antonfirsov:
@JimBobSquarePants replacing `IBufferManager` with `MemoryAllocator` is actually easy for me, what I can't figure out alone is to find the best way to acquire/manage the allocator instance.

I think we need to use `ImageSharpMiddlewareOptions.Configuration.MemoryAllocator` for allocation purposes by default. The question is that:
- Should we always use that insance (accessing it through `IOptions<ImageSharpMiddlewareOptions>`)? 
- Or does it still make sense to preserve `ImageSharpCoreBuilderExtensions.SetBufferManager()` according to ASP.Net Core best practices for DI?

JimBobSquarePants:
Interesting question. It depends really on whether our current allocation rules cover what we need for compressed streams. If you think that’s the case simply delete the service and we’ll use the memory allocator from the config.

antonfirsov:
@JimBobSquarePants looking at the shape of the current caching API made me think a bit, so I opened #32.

JimBobSquarePants:
Looking good! 

Changed `UseMemoryAllocatorFromMiddlewareOptions()` to `SetMemoryAllocatorFromMiddlewareOptions()` to make it more inline with other method names in the extensions. I also introduced an extension for adding it via a generic parameter.

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
```diff
@@ -62,15 +62,20 @@ public class PhysicalFileSystemCache : IImageCache
...
         /// Initializes a new instance of the <see cref="PhysicalFileSystemCache"/> class.
         /// </summary>
-        /// <param name="environment">The hosting environment the application is running in</param>
-        /// <param name="bufferManager">An <see cref="IBufferManager"/> instance used to allocate arrays transporting encoded image data</param>
-        /// <param name="options">The middleware configuration options</param>
-        public PhysicalFileSystemCache(IHostingEnvironment environment, IBufferManager bufferManager, IOptions<ImageSharpMiddlewareOptions> options)
+        /// <param name="environment">The hosting environment the application is running in.</param>
+        /// <param name="bufferManager">An <see cref="MemoryAllocator"/> instance used to allocate arrays transporting encoded image data.</param>
+        /// <param name="options">The middleware configuration options.</param>
+        public PhysicalFileSystemCache(IHostingEnvironment environment, MemoryAllocator bufferManager, IOptions<ImageSharpMiddlewareOptions> options)
```
antonfirsov: @JimBobSquarePants I forgot to rename the parameter `bufferManager` on a plenty of places.

JimBobSquarePants: Now worries, I'll fix that.

JimBobSquarePants:
@antonfirsov Can you have another look at this now please when you can? I've had a go at adding your `ICachedImage` concept.

src/ImageSharp.Web/Caching/IImageCache.cs
```diff
@@ -23,29 +23,30 @@ public interface IImageCache
         /// <summary>
         /// Gets the value associated with the specified key.
         /// </summary>
-        /// <param name="key">The cache key</param>
-        /// <returns>The <see cref="Task{IByteBuffer}"/></returns>
-        Task<IByteBuffer> GetAsync(string key);
+        /// <param name="key">The cache key.</param>
+        /// <returns>The <see cref="Task{IImageCache}"/>.</returns>
+        Task<ICachedImage> GetAsync(string key);
```
antonfirsov: I wonder if we need to keep the getting operation async. 

Maybe we should make the `Open*****()` operations async intead, but only if retrieving a Redis or Azure blob store stream is an asynchronous operation.

src/ImageSharp.Web/ICachedImage.cs
```diff
@@ -0,0 +1,32 @@
...
+
+using System;
+using System.IO;
+
+namespace SixLabors.ImageSharp.Web
+{
+    /// <summary>
+    /// Specifies the contract for managing image buffers from different locations.
+    /// </summary>
+    public interface ICachedImage
```
antonfirsov: alternative name ideas:
`IImageStreamSource` or simply `IImageSource`

src/ImageSharp.Web/Caching/IImageCache.cs
JimBobSquarePants: I did think about this actually. It depends basically on when users are likely to get the `LastWriteTimeUtc` property. I think it should always be  lazily in the `ICachedImage` or whatever we call it.

src/ImageSharp.Web/ICachedImage.cs
JimBobSquarePants: I think `IImageResolver` should become `IImageSource` and `ICachedImage` should become `IImageResolver`

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
```diff
@@ -170,7 +130,7 @@ public async Task<DateTimeOffset> SetAsync(string key, IByteBuffer value)
 
             using (FileStream fileStream = File.Create(path))
             {
-                await fileStream.WriteAsync(value.Array, 0, value.Length);
+                await stream.CopyToAsync(fileStream, (int)stream.Length).ConfigureAwait(false);
```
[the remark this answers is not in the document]
dlemstra: But the argument is not the size of the total buffer but the size of the internal copy buffer?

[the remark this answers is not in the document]
JimBobSquarePants: Ah, yeah, good spot. I misread the method. That buffer is crazy.

JimBobSquarePants:
Thanks @dlemstra for the additional review, very much appreciated! 👍 

@sebastienros It might be an idea for you to have a look at the changes we've been making here. I don't want to make things difficult/impossible for you guys downstream.

src/ImageSharp.Web/Caching/IImageCache.cs
```diff
@@ -21,31 +22,32 @@ public interface IImageCache
...
         /// <summary>
-        /// Gets the value associated with the specified key.
+        /// Gets the image resolver associated with the specified key.
         /// </summary>
-        /// <param name="key">The cache key</param>
-        /// <returns>The <see cref="Task{IByteBuffer}"/></returns>
-        Task<IByteBuffer> GetAsync(string key);
+        /// <param name="key">The cache key.</param>
+        /// <returns>The <see cref="IImageResolver"/>.</returns>
+        IImageResolver Get(string key);
```
antonfirsov: How about naming this `Lookup`?

src/ImageSharp.Web/Caching/IImageCache.cs
```diff
@@ -21,31 +22,32 @@ public interface IImageCache
...
         /// Returns the date and time, in coordinated universal time (UTC), that the value was last written to.
         /// </summary>
-        /// <param name="key">The cache key</param>
-        /// <param name="value">The value to store</param>
-        /// <returns>The <see cref="Task{DateTimeOffset}"/></returns>
-        Task<DateTimeOffset> SetAsync(string key, IByteBuffer value);
+        /// <param name="key">The cache key.</param>
+        /// <param name="stream">The stream containing the image to store.</param>
+        /// <returns>The <see cref="Task{DateTimeOffset}"/>.</returns>
+        Task<DateTimeOffset> SetAsync(string key, Stream stream);
```
antonfirsov: `StoreAsync`?

JimBobSquarePants: Get/Set are fairly standard. The NetFrameworks `ObjectCache` for example uses it.

JimBobSquarePants:
We now have a pooled `ChunkedMemoryStream` to deal with outStream allocations. 

This means removing `OpenWrite` from `IImageResolver` as we don't need it and we want to control the outstream type.

JimBobSquarePants:
I'm going to merge this in now so we can continue to work on #34

PR #34, opened 2018-08-30
Lock-free implementation of AsyncKeyLock

### LINKED ISSUE #31: AsyncKeyLock should use a ConcurrentDictionary
Gut feeling, but it looks like the hard lock is less efficient than what a ConcurrentDictionary would do. I ran https://github.com/benaadams/Ben.BlockingDetector (/cc @benaadams) and it showed up in the logs.

sebastienros:
Fixes #31

### Prerequisites


sebastienros:
There is a sample app in the repos, I will try to throw some load at it on our perf infra and see if it has any effect before you merge. I might add some little change to it (you'll judge) so I can automate it. It will be clear when you watch the https://live.asp.net video from last Tuesday.

sebastienros:
The result is that the perf is exactly the same, but it's not super good perf:

```
RequestsPerSecond:           5,351
Max CPU (%):                 23
WorkingSet (MB):             366
Avg. Latency (ms):           47.8
```

On the same machine we get from 20K (win) to 45K (lin) RPS for 16KB files (representative of the test I did). This is based on 256 concurrent connections.

I will get a perfview trace to see where most the time is spent in, but this PR doesn't have any impact at all. The fact that CPU is at 23% usually shows there is locking contention, so probably somewhere else.

sebastienros:
I look at the trace and saw `AsyncKeyLock` was the main element, and discovered that it was actually locking even on the fast path when a file is already processed and available in cache. So for each final image you can only have one request served at a time. I was not aware of that. I created another branch that changed the `AsyncKeyLock` to have `ReaderLockAsync()` and `WriterLockAsync()`. I got this on the same machine after:

```
RequestsPerSecond:           12,869
Max CPU (%):                 55
WorkingSet (MB):             334
Avg. Latency (ms):           19.8
```

It's much better, but in theory it should be faster than standard static files serving as the buffer is cached.
And I am not sure what is you really want to change this behavior so I am not creating a PR for now.

JimBobSquarePants:
Hi @sebastienros that excellent info thanks.

Basically the `AsyncKeyLock` exists to prevent file access errors. The cache API is designed in such a way that it can detect changes in the source files and will update the cached file (name based on the url + commands). Thus the following scenario can occur.

- An image is created and cached from a source + command.
- A subsequent request with the same details however the cache details that the source has been updated.
- A new cached image is created with the same name as the previous one but from the updated source During the time this is being written to disk (or wherever) another requests for the same source + command is received. 

So... What we want to avoid is the file contention that occurs when the cache attempts to read a cached file that is getting written.

Were your benchmarks against  my updated API #30 or the current stack? It should be faster in the new API.

I'd love to think of something that works faster but this was the best I could come up with. 

P.S I plan on watching your video today. Looks really interesting!!

JimBobSquarePants:
@sebastienros I've updated the PR with the latest codebase to make things easier. Once we get the throughput sorted I'll release another beta and help you get it wired up in Orchard.

sebastienros:
Good, I am going to run it again, thanks.
Also, with the changes I did on the sample project (nothing special, just closer to a default app) I would be able to run it continuously on our internal benchmarks infra. I'll let you know when it's available publicly, if you wish to be able to follow the results.

sebastienros:
It's worse with the new merge:

### Before

```
RequestsPerSecond:           5,206
Max CPU (%):                 21
WorkingSet (MB):             302
Avg. Latency (ms):           49
```

### After

```
RequestsPerSecond:           5,018
Max CPU (%):                 20
WorkingSet (MB):             278
Avg. Latency (ms):           50.9
```

Which means even more locking contention.


> So... What we want to avoid is the file contention that occurs when the cache attempts to read a cached file that is getting written.

I understand but right now it's locking all reads too. This is what my other branch is fixing, it's locking the reads when a write is being done. However I think it should still return the cached content while the file is written, and only update the cached content when it's done.

JimBobSquarePants:
Those timings are a surprise, we are definitely doing less work than we were before.

>I understand but right now it's locking all reads too. This is what my other branch is fixing, it's locking the reads when a write is being done. However I think it should still return the cached content while the file is written, and only update the cached content when it's done.

I agree. If we can optimize the process to only lock when processing image and writing the file that would be fine by me. I don't plan to add any cache cleanup code to the library so we should be safe.

sebastienros:
I am fine with it if you like it. At some point I will see how to make things much simpler at the cache level, but we already talked about it. For now the added perf will be sufficient for me.

JimBobSquarePants:
Thanks, Yeah would like to improve cache performance. Generating a key is too slow and intensive even for my liking.

PR #48, opened 2018-12-29
Replace IImageProvider.Get() with GetAsync()

kroymann:
### Prerequisites

### Description
This makes it possible for implementations of IImageProvider that rely on remote image sources to asynchronously create an IImageResolver.  This is important as it allows them to determine whether the requested image exists or not in an asynchronous manner prior to returning the IImageResolver.

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -170,7 +170,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
 
             bool processRequest = true;
             var imageContext = new ImageContext(context, this.options);
-            IImageResolver resolvedImage = provider.Get(context);
+            IImageResolver resolvedImage = await provider.GetAsync(context).ConfigureAwait(false);
 
             if (resolvedImage == null)
             {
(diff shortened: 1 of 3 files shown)


sebastienros:
Interesting, I think this was changed in the last beta, and the async logic is actually in the `IImageResolver` implementation, such that this method should not need async, or at least that's how I understood it when I adapted the code.

kroymann:
The issue with having the async logic in the IImageResolver and not in the IImageProvider is that you have to determine whether or not the source image exists in the provider _prior_ to creating the resolver.  If the source image doesn't exist, then the IImageProvider.Get() method needs to return null.  If instead you speculatively return an IImageResolver that later determines that the source image doesn't exist, there's no way to deal with it at that point.  Attempting to return null from the IImageResolver.OpenReadAsync() method results in a NullReferenceException.  I had a quick discussion about this with @JimBobSquarePants on [gitter](https://gitter.im/ImageSharp/General) and he agreed that this change seemed like the simplest and most logical solution to the problem.

sebastienros:
I am not disagreeing. and If the boss said so then it's all good.

JimBobSquarePants:
Yeah, oversight on my part. No way to check if the file exists asynchronously in Azure for example. I'm gonna merge this in, and then build a separate project with an Azure blob resolver to make sure it's all sensible before shipping another beta.

JimBobSquarePants:
Very useful, thanks!

I've started putting together an official version based on your code. There's a few performance fixes in there plus access control settings

https://github.com/SixLabors/ImageSharp.Web/tree/2d784e65f445ff493b28870de8c1097ee648e8d0/src/ImageSharp.Web.Providers.Azure

PR #49, opened 2018-12-29
Azure Blob Storage Providers

### LINKED ISSUE #25: Created a basic Azure Blob ImageResolver
Not sure of a better place to post this but here is a basic Azure Blob IImageResolver. 

https://gist.github.com/jakenuts/8e05fbd275e060775b6999bdf6d60c22

Thanks!

James

JimBobSquarePants:
### Prerequisites

### Description
Fixes #25 
- Adds a new library `SixLabors.ImageSharp.Web.Providers.Azure` that allows the retrieval of images from Azure Blob Storage.
- Simplifies setup configuration allowing replacement of individual parts.


src/ImageSharp.Web.Providers.Azure/Providers/BlobStorageImageProvider.cs
```diff
@@ -0,0 +1,102 @@
...
+        {
+            Guard.NotNull(options, nameof(options));
+            Guard.NotNull(storageOptions, nameof(storageOptions));
+
+            this.options = options.Value;
+            this.storageOptions = storageOptions.Value;
+            this.formatHelper = new FormatHelper(this.options.Configuration);
+            this.storageAccount = CloudStorageAccount.Parse(this.storageOptions.ConnectionString);
+
+            CloudBlobClient client = this.storageAccount.CreateCloudBlobClient();
```
kroymann: Have you tested (or otherwise confirmed) that the CloudBlobClient and CloudBlobContainer classes are thread safe?  The Azure Storage documentation is extremely vague on this topic (see [here](https://github.com/Azure/azure-storage-net/issues/732)) and it wasn't clear to me whether they can be created once here and then re-used in parallel by multiple different requests, or if they should be created per-request.

JimBobSquarePants: I've used a static `CloudBlobClient` before in ImageProcessor. Creating on per request will likely exhaust sockets. I'm confident that since we are not modifying the actual client we are safe.

JimBobSquarePants:
Going ahead with the merge. I'm very happy with how this has worked out.

PR #50, opened 2018-12-30
WIP: Add image cache metadata

kroymann:
_This is still a work-in-progress.  I'm creating a PR now so we can discuss the design and make sure this is a reasonable direction to pursue._

### Prerequisites

### Description
This adds a mechanism to store additional metadata alongside cached images.  The first use of this metadata cache is to store the Content-Type of the image so that we don't have to rely on making assumptions based on the filename, but can instead use the _actual_ image format information obtained when originally processing the image.

### DIFF
--- a/src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
+++ b/src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
@@ -9,6 +9,7 @@ using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Options;
+using SixLabors.ImageSharp.Web.Helpers;
 using SixLabors.ImageSharp.Web.Middleware;
 using SixLabors.ImageSharp.Web.Resolvers;
 using SixLabors.Memory;
@@ -36,6 +37,16 @@ namespace SixLabors.ImageSharp.Web.Caching
         public const string DefaultCheckSourceChanged = "false";
 
         /// <summary>
+        /// Filename extension for the metadata files.
+        /// </summary>
+        private const string MetaFileExtension = ".meta";
+
+        /// <summary>
+        /// Key for the Content-Type value in the metadata files.
+        /// </summary>
+        private const string ContentTypeKey = "Content-Type";
+
+        /// <summary>
         /// The hosting environment the application is running in.
         /// </summary>
         private readonly IHostingEnvironment environment;
@@ -54,6 +65,11 @@ namespace SixLabors.ImageSharp.Web.Caching
         /// The middleware configuration options.
         /// </summary>
         private readonly ImageSharpMiddlewareOptions options;
+
+        /// <summary>
+        /// Contains various helper methods based on the current configuration.
+        /// </summary>
+        private readonly FormatHelper formatHelper;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PhysicalFileSystemCache"/> class.
@@ -76,6 +92,7 @@ namespace SixLabors.ImageSharp.Web.Caching
             this.fileProvider = this.environment.WebRootFileProvider;
             this.memoryAllocator = memoryAllocator;
             this.options = options.Value;
+            this.formatHelper = new FormatHelper(this.options.Configuration);
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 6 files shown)


src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
```diff
@@ -146,5 +196,40 @@ public async Task<DateTimeOffset> SetAsync(string key, Stream stream)
         /// <param name="key">The cache key.</param>
         /// <returns>The <see cref="string"/>.</returns>
         private string ToFilePath(string key) => $"{this.Settings[Folder]}/{string.Join("/", key.Substring(0, (int)this.options.CachedNameLength).ToCharArray())}/{key}";
+
+        private async Task<Dictionary<string, string>> ReadMetadataFile(Stream stream)
```
JimBobSquarePants: This would have to be implementation agnostic. 

I'm thinking we create a custom struct that can be easily serialized deserialized into raw bytes that can be used for storing both the content type and last modified date. 

CacheMeta
8 bytes | ModifiedDate
2 bytes | Content Type string length
n bytes | Content Type

kroymann: My thinking with this implementation was to use a file format that provides easy forward/backward compatibility so that additional metadata fields could be easily added in the future (I already have one such use that I would like to discuss if you end up deciding to proceed with this metadata concept).  I would have used JSON, but didn't want to take a dependency on a JSON library, so I went with a really simple text-based key/value pair file format.  I wasn't too worried about wasting a few extra bytes by having the data stored as text as the size should be trivial relative to the actual image data that it is annotating.

kroymann: But thinking about it now, I do like the idea of having the metadata contained in a struct, as that would make it possible to add new metadata fields without needing to add any additional methods or change function signatures.  Something like:

```
public struct ImageMetadata
{
    public DateTime LastModifiedUtc;
    public string ContentType;
}

public interface IImageResolver
{
    Task<Stream> OpenReadAsync();
    Task<ImageMetadata> GetImageMetadata();
}

public interface IImageCache
{
    Task<IImageResolver> GetAsync(string key);
    Task<DateTimeOffset> SetAsync(string key, Stream stream, ImageMetadata metadata);
}
```

kroymann:
I pushed another commit that refactors the ImageMetadata into it's own simple class.  For now I kept the simple text-based key/value pair serialization format as it has the benefit of being forward/backward compatible.

One thing to note: While making this change I shuffled around the logic for dealing with expired cache entries.  The IImageCache is now purely responsible for generating IImageResolver instances, which in turn expose ImageMetadata containing the LastModified timestamp, and the ImageSharpMiddleware is now responsible for comparing timestamps and determining whether it can use a cached image or not.

JimBobSquarePants:
I like the expires changes and the overall shape. I've got some ideas though and am gonna push them directly to your PR. Hopefully you'll see where I'm going here.

JimBobSquarePants:
@kroymann Cached response type and extension is now solely the responsibility of the cache. I think this is starting to look good now.

I've gone ahead with the byte aligned struct approach to reduce `ImageMetaData` file size (we really need this to be as small as possible) we're maxing out at 18 bytes now. 

I did notice that I was getting an odd AccessViolationException in `ImageMetaData` when moving between debug and release mode but I'm assuming that must be something to do with a bug in the CLR somewhere as deleting the cache seemed to fix it. Would certainly like a second opinion though. 

@sebastienros What do you think of all this. It certainly seems more robust than before, should be much faster too on network based providers/caches.

JimBobSquarePants:
Bugger. `System.AccessViolationException` looks like the real deal now, I'm getting failed tests on Travis too. I'll try and figure out what I broke in the last commit.

JimBobSquarePants:
Ok... Symptoms are as follows.

If the cache is empty I can start the sample application and it will generate images plus metadata. ctrl+F5 works no problem. If I then close the tab and rebuild the application loading all images will fail with a `AccessViolationException` when parsing the metadata file stream. I'm stumped for now as to why and it's 2am so gonna stop.

kroymann:
@JimBobSquarePants I'm curious about your rationale for wanting to serialize the metadata into the smallest possible packed binary format.  Doing so causes the file format to become tightly coupled to the version of the code which means that if we want to add a new field in the future (and I already have such a use case in mind to discuss after we sort this out), the parsing code will get increasingly complex unless we're going to simply ignore files that were cached using previous versions of the file metadata file format.  Using a format like Json, or ProtoBuf, or the simple key/value-pair format from the beginning of this PR, the file becomes both forward and backward compatible.  And relative to the file size of the image data itself, we're talking about a miniscule difference in disk space usage, right?

JimBobSquarePants:
@kroymann You got me there, that's a damn fine idea which I have implemented. 

I think this is ready to go so please have a good dig through to see if I have missed anything.

kroymann:
@JimBobSquarePants I reviewed the cache-control change.  Looks exactly like what I was envisioning.  I gave everything a second read through and it all looks good to me.

Issue #56, opened 2019-01-31
Race condition can lead to dead-lock in AsyncKeyLock


kroymann:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp.Web
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I was reviewing the latest AsyncKeyLock/Doorman implementation and have found a race condition that could cause a caller to be blocked forever waiting to receive the lock, or could potentially cause a reader and writer to execute concurrently.  Here are the problematic flows:

**How to get permanently stuck waiting for the lock**
1. Thread A begins releasing a **writer** lock (with no waiting readers or writers)
2. Thread B attempts to obtain the lock (and gets stuck waiting to lock the Doorman.waitingWriters queue)
3. Thread A finishes releasing the writer lock. It leaves the Doorman.status field set to -1, and removes the Doorman from the ConcurrentDictionary as it believes nobody is waiting.
5. Thread B continues, observes that the Doorman.status field is -1, and then adds itself to a waiting queue that will never be released.

**How to cause a reader and writer to execute concurrently on the same key**
1. Thread A begins releasing a **reader** lock (again with no waiting readers or writers)
2. Thread B attempts to obtain the **writer** lock
3. Thread A finishes releasing the reader lock, and incorrectly removes the lock from the ConcurrentDictionary as it believes nobody is waiting.
4. Thread B continues and observes that the Doorman status is 0 so it takes the writer lock and proceeds.
5. Thread C attempts to take **reader** lock. Since the Doorman in use by thread B is no longer stored in the ConcurrentDictionary, thread C creates creates a new Doorman using the same key.  Now Thread B and Thread C are holding two different locks for the same key.

### Dicussion
There are two problems here:
1. The status field is not getting set back to 0 when the final writer releases the lock.
2. The act of obtaining the Doorman from the dictionary and then adding a reference to that Doorman (thus preventing it from being removed from the dictionary) must be atomic.  I've been trying to think of a way to do this using lock-free programming and have not yet come up with a solution.  Unfortunately, I think this will require going back to a locking paradigm.

PR #57, opened 2019-01-31
Fix race-condition in AsyncKeyLock

kroymann:
### Prerequisites

### Description
This change addresses the race conditions identified in #56 by adding a RefCount property to the Doorman class and then using a SpinLock to serialize access to the Keys dictionary.  It also adds a pool so that unused doormen can be re-used in order to cut down on GC allocations and reduce the time spent holding the SpinLock.

### DIFF
--- a/src/ImageSharp.Web/Caching/AsyncKeyLock.cs
+++ b/src/ImageSharp.Web/Caching/AsyncKeyLock.cs
@@ -2,8 +2,8 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System;
-using System.Collections.Concurrent;
-using System.Runtime.CompilerServices;
+using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Thanks for looking into this @kroymann This may explain the issues I had when using `Thread.Sleep` over `Task.Delay` in the unit tests. I'll have a proper look asap.

https://github.com/SixLabors/ImageSharp.Web/blob/82929ea0f4f202036dd96ac694af7d78e5860452/tests/ImageSharp.Web.Tests/Caching/AsyncKeyLockTests.cs#L66

@sebastienros Can you also cast your eye over this when you can.

JimBobSquarePants:
Did some localised testing with your changes and it all looks to be good to me. Great work! 👍

PR #69, opened 2019-08-15
Improve ImageSharpMiddleware no-op case's performance

lahma:
### Prerequisites

### Description

I profiled [Orchard Core](https://github.com/OrchardCMS/OrchardCore) as @sebastienros gave the idea. I saw that ImageSharp was doing some unnecessary processing as part of the pipeline when the request ended to be no-op on ImageSharp's behalf. I tweaked following bits:

* reduce usage of LINQ, which is always more expensive than "code nearer to the metal"
* change error handling to be more easy to inline (micro optimization)
* use concrete types as fields, no extra penalty from interface access
* split the actual work to `ProcessRequest` when actually needed
* don't parse in `OnParseCommands` when result is discarded

### DIFF
--- a/src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
+++ b/src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.WebUtilities;
+using Microsoft.Extensions.Primitives;
 
 namespace SixLabors.ImageSharp.Web.Commands
 {
@@ -21,7 +22,7 @@ namespace SixLabors.ImageSharp.Web.Commands
                 return new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             }
 
-            var parsed = QueryHelpers.ParseQuery(context.Request.QueryString.ToUriComponent());
+            Dictionary<string, StringValues> parsed = QueryHelpers.ParseQuery(context.Request.QueryString.ToUriComponent());
             var transformed = new Dictionary<string, string>(parsed.Count);
             foreach (var pair in parsed)
             {
--- a/src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
+++ b/src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.WebUtilities;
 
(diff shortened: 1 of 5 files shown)


src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
```diff
@@ -12,18 +11,24 @@ namespace SixLabors.ImageSharp.Web.Commands
...
         {
-            if (!context.Request.Query.Any())
+            if (context.Request.Query.Count == 0)
             {
                 return new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
             }
 
-            return QueryHelpers.ParseQuery(context.Request.QueryString.ToUriComponent())
-                               .ToDictionary(k => k.Key, v => v.Value.ToString());
+            var parsed = QueryHelpers.ParseQuery(context.Request.QueryString.ToUriComponent());
```
JimBobSquarePants: It's hard to tell what the returned type of `parsed` is here, can we use the explicit declaration.

src/ImageSharp.Web/Guard.cs
```diff
@@ -23,18 +24,26 @@ internal static class Guard
...
-                {
-                    throw new ArgumentNullException(parameterName, message);
-                }
+                ThrowArgumentNull<T>(parameterName, message);
+            }
+        }
 
-                throw new ArgumentNullException(parameterName);
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static void ThrowArgumentNull<T>(string parameterName, string message)
```
JimBobSquarePants: We can reduce codegen here by dropping the generic type parameter.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -149,14 +152,30 @@ public class ImageSharpMiddleware
         /// <returns>The <see cref="Task"/>.</returns>
         public async Task Invoke(HttpContext context)
         {
-            IDictionary<string, string> commands = this.requestParser.ParseRequestCommands(context)
-                .Where(kvp => this.knownCommands.Contains(kvp.Key))
-                .ToDictionary(p => p.Key, p => p.Value);
+            var commands = this.requestParser.ParseRequestCommands(context);
```
JimBobSquarePants: It's hard to tell what the returned type of `commands` is here, can we use the explicit declaration

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -165,21 +184,32 @@ public async Task Invoke(HttpContext context)
...
                 // Log the error but let the pipeline handle the 404
+                var imageContext = new ImageContext(context, this.options);
                 this.logger.LogImageResolveFailed(imageContext.GetDisplayUrl());
                 processRequest = false;
             }
 
+            if (!processRequest)
+            {
+                // Call the next delegate/middleware in the pipeline
+                await this.next(context).ConfigureAwait(false);
```
JimBobSquarePants: We need to `return` after this line.

lahma:
Thank you for the review, good points! I hope I got the fixes right and PR should be good now.

JimBobSquarePants:
Perfect! I'll merge this now, thanks for your help!

lahma:
You're most welcome,  thanks for accepting the PR!

PR #71, opened 2019-08-24
Suggested change to logic of command processing

deanmarcussen:
### Prerequisites

### Description

This is a suggested change to the logic of when commands are parsed, to further improve the no-op performance.

ImageSharp no-op's based on the result from an `IImageProvider`, which is only passed the `HttpContext` not the commands that have been processed by the request parser.

The `commands` are not used until the images are processed, as far as I can tell, unless I'm missing something?

I don't think the `OnParseCommands` Action can be used to affect the validation of a request - that functionality is with the `IImageProvider`.

With this change to the logic, `OnParseCommands` can still perform mutations to the commands, before processing.

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -152,20 +152,6 @@ namespace SixLabors.ImageSharp.Web.Middleware
         /// <returns>The <see cref="Task"/>.</returns>
         public async Task Invoke(HttpContext context)
         {
-            IDictionary<string, string> commands = this.requestParser.ParseRequestCommands(context);
-            if (commands.Count > 0)
-            {
-                foreach (var command in commands.Keys)
-                {
-                    if (!this.knownCommands.Contains(command))
-                    {
-                        commands.Remove(command);
-                    }
-                }
-            }
-
-            this.options.OnParseCommands?.Invoke(new ImageCommandContext(context, commands, CommandParser.Instance));
-
             // Get the correct service for the request.
             IImageProvider provider = null;
             foreach (var resolver in this.resolvers)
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
Thanks @deanmarcussen  but `OnParseCommands` can actually be used to validate a request, it limits upper limit resize dimensions by default.

https://github.com/SixLabors/ImageSharp.Web/blob/b02a6f7f9640507858647fa4b66a7f9753fc0e5e/src/ImageSharp.Web/Middleware/ImageSharpMiddlewareOptions.cs#L41

What we should have though is another check for the commands length after it is called and return when there are none to run. I've opened PR #73 to address this.

deanmarcussen:
Ok, cool, cheers for looking at it.

So with #73 now returning on no commands (which works for me!), is the `IRequestParser` considered the appropriate place to validate things like the `RequestPath` ?

For example, if all media is always accessed through the `/media` path we could now validate this in a custom RequestParser which would return an empty dictionary, which would cause a no-op, rather than waiting to do this in `IsValidRequest` from the provider, which will happen sometime later after all the commands have been parsed (which we don't need, because we'll end up no-opping).

Just asking because I'd like to be sure we're working this as intended 😄 

As a side bonus, #73 will stop ImageSharp processing images that don't need resizing, something we've been stopping in the `IsValidRequest` of the provider.

JimBobSquarePants:
`IRequestParser` is designed primarily for converting incoming requests into a series of commands, not for validation. 

`IImagProvider.Match` is for matching things like the request path, basically if the path does not start with `media` you would ignore it. See the [AzureImageProvider](https://github.com/SixLabors/ImageSharp.Web/blob/5ced07bb255f56b62f2c819d161c17e482e72c00/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs#L92) example 

`IImagProvider.IsValidRequest` is for additional validation. e.g does it have an image extension?

All are customizable of course.

PR #76, opened 2019-09-09
Prevent concurrent modification when iterating request commands

lahma:
### Prerequisites

- [x]  I have written a descriptive pull-request title

### Description

There was an unfortunate problem in optimization PR which surfaced in OrchardCore. Commands were mutated with active iterator and that broke the contract. This fixes the problem and hopefully you could do yet another release to NuGet 🙏

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -157,7 +157,8 @@ namespace SixLabors.ImageSharp.Web.Middleware
             IDictionary<string, string> commands = this.requestParser.ParseRequestCommands(context);
             if (commands.Count > 0)
             {
-                foreach (string command in commands.Keys)
+                List<string> commandKeys = new List<string>(commands.Keys);
+                foreach (string command in commandKeys)
                 {
                     if (!this.knownCommands.Contains(command))
                     {


src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -157,7 +157,8 @@ public async Task Invoke(HttpContext context)
             IDictionary<string, string> commands = this.requestParser.ParseRequestCommands(context);
             if (commands.Count > 0)
             {
-                foreach (string command in commands.Keys)
+                List<string> commandKeys = new List<string>(commands.Keys);
```
JimBobSquarePants: I would have used `ToArray()` here but a list will do for now.

JimBobSquarePants:
@lahma This is certainly critical enough to warrant a new release so here you go! 😄 

https://www.nuget.org/packages/SixLabors.ImageSharp.Web/1.0.0-beta0009

lahma:
Thanks for the prompt action and release (and sorry for the extra chore)! 👍

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
lahma: I dread to use System.Linq in active/hot path 😉

Issue #78, opened 2019-09-23
cache path isn't valid


yohsii:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Web
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
ImageSharp.Web is no longer working in asp.net 3.0 (full release version). the error is:
```
ArgumentException: The path in 'value' must start with '/'. (Parameter 'value')
Microsoft.AspNetCore.Http.PathString..ctor(string value)
Microsoft.AspNetCore.Hosting.StaticWebAssets.StaticWebAssetsFileProvider.StartsWithBasePath(string subpath, out PathString rest)
Microsoft.AspNetCore.Hosting.StaticWebAssets.StaticWebAssetsFileProvider.GetFileInfo(string subpath)
Microsoft.Extensions.FileProviders.CompositeFileProvider.GetFileInfo(string subpath)
SixLabors.ImageSharp.Web.Caching.PhysicalFileSystemCache.GetAsync(string key)
SixLabors.ImageSharp.Web.Middleware.ImageSharpMiddleware.ProcessRequestAsync(HttpContext context, bool processRequest, IImageResolver sourceImageResolver, ImageContext imageContext, IDictionary<string, string> commands)
SixLabors.ImageSharp.Web.Middleware.ImageSharpMiddleware.Invoke(HttpContext context)
Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.MigrationsEndPointMiddleware.Invoke(HttpContext context)
Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.Invoke(HttpContext httpContext)
Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore.DatabaseErrorPageMiddleware.Invoke(HttpContext httpContext)
Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware.Invoke(HttpContext context)
```
the fix for this issue is straightforward, in PhysicalFileSystemCache.cs on line 75 there's a call:
```
string path = this.ToFilePath(key);
```

[...]

PR #79, opened 2019-09-30
added forward slash to path as it's being passed to PathString constructor which expects the path to start with /

yohsii:
addresses issue #78. seems to be asp.net core 3.0 related, hopefully adding the slash doesn't break earlier releases of asp.net core.

### Prerequisites

- [x ] I have written a descriptive pull-request title
- [x ] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp.Web/pulls) open


JimBobSquarePants:
Thanks for this. Looks like it's breaking on Linux now though.

JimBobSquarePants:
I'm closing this as it was fixed by #80

PR #80, opened 2019-10-10
Create PhysicalFileProvider from Webroot

deanmarcussen:
# Prerequisites

### Description

Fixes https://github.com/SixLabors/ImageSharp.Web/issues/78

I've updated `PhysicalFileSystemCache` to create a `PhysicalFileProvider` from the `IHostingEnvironment.WebRootPath` to work around the issues some people are experiencing in development mode with core 3

It has meant taking a reference on `Microsoft.Extensions.FileProviders.Physical` but I think I've got the correct versions for both targets `netstandard1.4` and `netstandard2.0`, noting there is no released version of 1.1.2 which would match the other `netstandard1.4` targets


JimBobSquarePants:
@deanmarcussen Looks good to me, thanks!

agriffard:
Does it solve #4419 @deanmarcussen ?

PR #81, opened 2019-12-18
Check if file exists before matching in PhysicalFileSystemProvider

agrigg:
### Prerequisites

### Description
The Match method on the PhysicalFileSystemProvider always returns true. This creates a problem if you add the AzureBlobStorageImageProvider after the PhysicalFileSystemProvider because the AzureBlobStorageImageProvider will never be utilized.

I've added a Match method to the PhysicalFileSystemProvider that checks to see if the file exists before returning true.

### DIFF
--- a/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
+++ b/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
@@ -27,6 +27,11 @@ namespace SixLabors.ImageSharp.Web.Providers
         private readonly FormatUtilities formatUtilities;
 
         /// <summary>
+        /// A match function used by the resolver to identify itself as the correct resolver to use.
+        /// </summary>
+        private Func<HttpContext, bool> match;
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Hi @agrigg 

Thanks for this but the `Match` function is for matching URLs not for determining whether the file exists or not. By adding the file info check you would significantly hurt performance. 

There's actually an example in the tests project which demonstrates how to register both regardless of order. 

https://github.com/SixLabors/ImageSharp.Web/blob/5901fff5dc3d216817779d3ac1739fa78f7d86b1/tests/ImageSharp.Web.Tests/ImageSharpTestServer.cs#L54

agrigg:
@JimBobSquarePants Gotcha, thanks for pointing me to that example, that is really helpful. It would be nice to update the README to make a note of that. Would you want me to create a PR for that?

JimBobSquarePants:
@agrigg Yeah if you like, thanks! I do plan on documenting everything properly in this repository also so i can use what you add as a basis for that.

https://github.com/SixLabors/docs

PR #82, opened 2019-12-18
Allow use of any container in AzureBlobStorageImageProvider

agrigg:
### Prerequisites

### Description
The current AzureBlobStorageImageProvider only supports a single blob container. Many applications will want to utilize multiple containers. I've updated the AzureBlobStorageImageProvider to utilize a RoutePrefix and then to parse the container name and blob file name from the URL after the prefix.

So, an example URL with the corresponding parts would be:

https://localhost:9999/assets/products/productImage.jpg

Route Prefix: assets
Container Name: products
Blob Filename: productImage.jpg

The AzureBlobStorageImageProvider only matches on paths that start with the route prefix and then parses the URL to pull the container and blob file name.

I also removed the code for creating a new container if it doesn't exists because it seems like this should be a read-only operation and now actually adding containers in Azure.

### DIFF
--- a/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
+++ b/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System;
+using System.Linq;
 using System.Net;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
@@ -28,11 +29,6 @@ namespace SixLabors.ImageSharp.Web.Providers
         /// The cloud storage account.
         /// </summary>
         private readonly CloudStorageAccount storageAccount;
-
-        /// <summary>
-        /// The container in the blob service.
-        /// </summary>
-        private readonly CloudBlobContainer container;
 
         /// <summary>
         /// The blob storage options.
@@ -63,25 +59,6 @@ namespace SixLabors.ImageSharp.Web.Providers
             this.storageOptions = storageOptions.Value;
             this.formatUtilities = formatUtilities;
             this.storageAccount = CloudStorageAccount.Parse(this.storageOptions.ConnectionString);
-
-            try
-            {
-                // It's ok to create a single reusable client since we are not altering it.
-                CloudBlobClient client = this.storageAccount.CreateCloudBlobClient();
-                this.container = client.GetContainerReference(this.storageOptions.ContainerName);
-
-                if (!this.container.Exists())
-                {
-                    this.container.Create();
-                    this.container.SetPermissions(new BlobContainerPermissions { PublicAccess = this.storageOptions.AccessType });
-                }
-            }
-            catch (StorageException storageException) when (storageException.RequestInformation.HttpStatusCode == (int)HttpStatusCode.Conflict
-                || storageException.RequestInformation.ExtendedErrorInformation.ErrorCode == StorageErrorCodeStrings.ContainerAlreadyExists)
-            {
-                // https://github.com/Azure/azure-sdk-for-net/issues/109
-                // We do not fire exception if container exists - there is no need in such actions
-            }
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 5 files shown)


JimBobSquarePants:
Hmmmm..... I'm not sure about this for a couple of reasons.

1. I don't want to have to deal with exceptions caused by a lack of container.
2. I don't see a significant business case for utilizing multiple containers over good naming conventions for nested blob folders.

agrigg:
@JimBobSquarePants Thanks for taking a look at the PR. In regard to your comments:

1. It shouldn't throw exceptions if a container doesn't exist because we check that the container exists and if it doesn't the code returns null like it does if the blob doesn't exist.

2. If you're starting a new application you could put everything in 1 container, but people with existing applications may already have multiple containers. Sometimes you may want to put items in different containers to separate customer data or to give different permissions per container. Having worked a good bit in ASP.NET and Azure I find it can be helpful to have several containers.

Suchiman:
@JimBobSquarePants another case for containers is that folders aren't real, they're just part of the name. So if you want to delete a folder, you need to list all blobs with that prefix (folder), then delete them individually, which affects billing too, while you can just say, delete the whole container.

JimBobSquarePants:
When would you delete the cache files?

Suchiman:
Depends on the usecase, when the original images are deleted or after a certain amount of time (i generally use monthly partitioning in blob storage for data that becomes irrelevant past a certain point), although, one could use the more or less recent addition of lifecycle management to clear out aged blobs.

JimBobSquarePants:
Ok @agrigg I've figured out what I don't like in the changes and think we can re-jig it slightly to make it work.

Basically. I want to drop the prefix and rely on registering the individual containers instead. I also don't want to create a container reference for each image to resolve, it's just too expensive an operation.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
```diff
@@ -29,11 +30,6 @@ public class AzureBlobStorageImageProvider : IImageProvider
         /// </summary>
         private readonly CloudStorageAccount storageAccount;
 
-        /// <summary>
-        /// The container in the blob service.
-        /// </summary>
-        private readonly CloudBlobContainer container;
```
JimBobSquarePants: Make this a `Dictionary<string, CloudBlobContainer>` which we create on startup.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
```diff
@@ -63,25 +59,6 @@ public class AzureBlobStorageImageProvider : IImageProvider
             this.storageOptions = storageOptions.Value;
             this.formatUtilities = formatUtilities;
             this.storageAccount = CloudStorageAccount.Parse(this.storageOptions.ConnectionString);
-
-            try
```
JimBobSquarePants: I'm ok dropping this. We should probably throw and exception if no container exists for the given name though.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
```diff
@@ -94,19 +71,24 @@ public class AzureBlobStorageImageProvider : IImageProvider
...
         {
-            // Strip the leading slash and container name from the HTTP request path and treat
-            // the remaining path string as the blob name.
-            // Path has already been correctly parsed before here.
-            string blobName = context.Request.Path.Value.TrimStart(SlashChars)
-                                     .Substring(this.storageOptions.ContainerName.Length)
-                                     .TrimStart(SlashChars);
-
-            if (string.IsNullOrWhiteSpace(blobName))
+            var parts = new AzureBlobStoragePathParts(context.Request.Path.Value, this.storageOptions.RoutePrefix);
```
JimBobSquarePants: I like the idea of a parts type. It should be a readonly struct though.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
```diff
@@ -94,19 +71,24 @@ public class AzureBlobStorageImageProvider : IImageProvider
...
+
+            if (string.IsNullOrEmpty(parts.BlobFilename))
+            {
+                return null;
+            }
+
+            CloudBlobClient client = this.storageAccount.CreateCloudBlobClient();
```
JimBobSquarePants: Create the client and containers at startup. The options should take a collection of strings representing the containers to register.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
```diff
@@ -122,7 +104,7 @@ public bool IsValidRequest(HttpContext context)
         private bool IsMatch(HttpContext context)
         {
             string path = context.Request.Path.Value.TrimStart(SlashChars);
-            return path.StartsWith(this.storageOptions.ContainerName, StringComparison.OrdinalIgnoreCase);
+            return path.StartsWith(this.storageOptions.RoutePrefix, StringComparison.OrdinalIgnoreCase);
```
JimBobSquarePants: Check the keys in the dictionary here for the container name.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProviderOptions.cs
```diff
@@ -11,20 +11,14 @@ namespace SixLabors.ImageSharp.Web.Providers
...
+        /// Gets or sets the route prefix for processing azure assets.
+        /// Url format would be: /{RoutePrefix}/{ContainerName}/{FileName}
         /// </summary>
-        public string ContainerName { get; set; }
-
-        /// <summary>
-        /// Gets or sets the value indicating the level of public access that is allowed on the container.
-        /// </summary>
-        public BlobContainerPublicAccessType AccessType { get; set; } = BlobContainerPublicAccessType.Blob;
+        public string RoutePrefix { get; set; }
```
JimBobSquarePants: Make this a collection of type string called `ContainerNames`.

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStoragePathParts.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace SixLabors.ImageSharp.Web.Providers
+{
+    /// <summary>
+    /// Components of Azure blob URL path
+    /// </summary>
+    public class AzureBlobStoragePathParts
```
JimBobSquarePants: We should make this a struct to reduce heap allocations.

JimBobSquarePants:
@agrigg  @Suchiman I've opened a new PR #91 to cover this use case. Please take a look to ensure it covers your requirements.

Issue #77, opened 2019-09-10
BlobStorageImageProvider won't work anymore without query parameters


wagich:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Web
- [ ] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
Since merging pull request #73 "Return on no commands", the `ImageSharpMiddleware` directly hands off request processing to the next middleware in the pipeline when there aren't any commands found.

This prevents any provider fetching images from a remote source (as the Azure Blob Storage Provider does) from running in a "pass-through" scenario without any transforms. Only images present on the filesystem still work, because the `StaticFileMiddleware` handles those requests.

It would be extremely unfortunate to need a second, separate middleware to handle those requests given that it would need to completely duplicate the functionality of the already present provider.

### Steps to Reproduce
- Default config with AzureBlobStorageImageProvider set as image provider
- Access image URL without any query parameters

### System Configuration
- ASP.NET Core 3.0 Preview 9
- SixLabors.ImageSharp.Web 1.0.0-dev000227

wagich:
On second thought, there could be a config flag to turn the early-return behavior off, that would keep the (I'm assuming) common case fast but still enable the pass-through scenario for custom providers.

JimBobSquarePants:
@wagich The system was designed originally to not process images without commands but I can see a purpose for that now with the new external image providers. A configurable switch does seem like a sensible approach 👍

wagich:
Would you be open to a PR? I'll need to implement it anyhow in my code and would be happy to contribute.

One alternative to a config switch could also be to move the switch to the provider. Then the middleware could bail (or continue) after chosing a provider.



On Fri, Sep 13, 2019 at 4:14 PM +0200, "James Jackson-South" <notifications@github.com<mailto:notifications@github.com>> wrote:


@wagich<https://github.com/wagich> The system was designed originally to not process images without commands but I can see a purpose for that now with the new external image providers. A configurable switch does seem like a sensible approach 👍

andymac4182:
This would be great! We ended up implementing our own middleware that just returned the raw image.

[...]

PR #83, opened 2020-01-15
Add a setting so providers can process images when no commands are supplied.

Pandorax100:
### Prerequisites

### Description

Addresses #77 

This exposes a setting on individual providers to determine whether they should continue processing an image when no commands are supplied. This is important for remote providers - such as the `AzureBlobStorageImageProvider` - as they are unable to fall back on the static files middleware to serve images.

### DIFF
--- a/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
+++ b/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
@@ -85,7 +85,7 @@ namespace SixLabors.ImageSharp.Web.Providers
         }
 
         /// <inheritdoc/>
-        public ProcessingBehavior ProcessingBehavior { get; set; } = ProcessingBehavior.All;
+        public ProcessingBehavior ProcessingBehavior { get; } = ProcessingBehavior.All;
 
         /// <inheritdoc/>
         public Func<HttpContext, bool> Match
--- a/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
+++ b/src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
@@ -85,7 +85,7 @@ namespace SixLabors.ImageSharp.Web.Providers
         }
 
         /// <inheritdoc/>
-        public bool ProcessWhenNoCommands { get; set; } = true;
+        public ProcessingBehavior ProcessingBehavior { get; set; } = ProcessingBehavior.All;
 
         /// <inheritdoc/>
         public Func<HttpContext, bool> Match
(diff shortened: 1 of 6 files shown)


src/ImageSharp.Web/Providers/IImageProvider.cs
```diff
@@ -14,6 +14,12 @@ namespace SixLabors.ImageSharp.Web.Providers
     /// </summary>
     public interface IImageProvider
     {
+        /// <summary>
+        /// Gets or sets a value indicating whether this provider should process
+        /// an image when no commands have been supplied.
+        /// </summary>
+        bool ProcessWhenNoCommands { get; set; }
```
JimBobSquarePants: This shouldn't be a settable property on the interface. We don't want it to be configurable during startup

I think we should rename the property also and make it an enum.

```c#
ProcessingBehavior  ProcessingBehavior { get; }
```

``` c#
public enum ProcessingBehavior
{
    CommandOnly,
    All
}
```

Pandorax100:
I completely agree. The interface has had the setter removed, and the setting has been changed to an enum.

deanmarcussen:
This looks good.

My only caveat (just from working around this issue) is that when an image comes in with no commands and is supported as `ProcessingBehaviour.All` the image will still get loaded by ImageSharp and saved back out as a different stream/image.

So it just means that the image will always be mutated in some way (which has an impact on memory / performance)

It would be nice if when there are no commands the image wasn't mutated, but I leave that choice with @JimBobSquarePants

src/ImageSharp.Web.Providers.Azure/Providers/AzureBlobStorageImageProvider.cs
```diff
@@ -85,7 +85,7 @@ public class AzureBlobStorageImageProvider : IImageProvider
         }
 
         /// <inheritdoc/>
-        public bool ProcessWhenNoCommands { get; set; } = true;
+        public ProcessingBehavior ProcessingBehavior { get; set; } = ProcessingBehavior.All;
```
JimBobSquarePants: Nearly there. Just need to drop the setters on the concrete implementations 👍

Pandorax100: Already done 👍  I realised I'd missed that in the initial commit so it's there in https://github.com/SixLabors/ImageSharp.Web/pull/83/commits/105a80eb9609637af9a9c36f18307356935987ae

JimBobSquarePants:
@deanmarcussen makes a valid point [above](https://github.com/SixLabors/ImageSharp.Web/pull/83#issuecomment-575888396)

@Pandorax100  We're still fully loading and saving the image when there are no commands. 
We still need to identify the format so when there are no commands we should use the `Image.Identify(stream, out IImageFormat format)` method to get the correct mimetype.

I'm happy to accept this as-is and do the work though. I'm currently updating the build process and changing the targets to sensible ones (Core 3.1, Core 2.1) so don't want to keep you waiting longer than you need to.

Pandorax100:
Yeah, I see the problem

I've just had a little look and it doesn't seem like `Image.Identify(stream, out IImageFormat format)` is available with the version of ImageSharp referenced by this project. I see that there is `Image.DetectFormat(inStream)` but if you're in the process of updating the build process etc then I think it's probably best to accept this as is and we can come back to it at a later date. If that's alright with you?

JimBobSquarePants:
Totally fine. Will get this merged ASAP.

PR #84, opened 2020-01-21
New build process

JimBobSquarePants:
### Prerequisites

### Description
- Adopts Github Actions as a build process
- Updates dependencies
- Move targets to sensible values.
- Optimize no-command pipline


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,105 @@
...
+
+            - name: Update Codecov
+              uses: iansu/codecov-action-node@v1.0.0
+              if: matrix.options.codecov == true
+              with:
+                  token: ${{secrets.CODECOV_TOKEN}}
+                  file: "coverage.${{matrix.options.framework}}.xml"
+                  flags: unittests
+
+            - name: Pack # We can use this filter as we know it happens only once and takes the most time to complete.
```
tocsoft: we should be using the split build/publish workflow here not skip and hope it takes the longest

GitVersion.yml
```diff
@@ -0,0 +1,6 @@
+continuous-delivery-fallback-tag: ci
+branches:
+    master:
+        tag: unstable
```
tocsoft: isn't this missing the `ContinuousDeployment` mode?

.github/workflows/build-and-test.yml
JimBobSquarePants: Yeah, good spot. I'll update that once I get it working.

GitVersion.yml
JimBobSquarePants: Cripes! Fixed

PR #89, opened 2020-03-16
Update shared infrastructure

brianpopow:
### Prerequisites

### Description

Updates the shared infrastructure of the ImageSharp.Web project

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ Core Team
 - [Dirk Lemstra](https://github.com/dlemstra)
 - [Anton Firsov](https://github.com/antonfirsov)
 - [Scott Williams](https://github.com/tocsoft)
+- [Brian Popow](https://github.com/brianpopow)
 
 ### Backers
 
--- a/tests/ImageSharp.Web.Tests/Commands/CommandParserTests.cs
+++ b/tests/ImageSharp.Web.Tests/Commands/CommandParserTests.cs
@@ -75,10 +75,12 @@ namespace SixLabors.ImageSharp.Web.Tests.Commands
 
         public static TheoryData<object, string> Rgba32Values = new TheoryData<object, string>
         {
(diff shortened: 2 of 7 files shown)


tests/ImageSharp.Web.Tests/Commands/CommandParserTests.cs
```diff
@@ -75,10 +75,12 @@ public class CommandParserTests
 
         public static TheoryData<object, string> Rgba32Values = new TheoryData<object, string>
         {
-            { Rgba32.White, "255,255,255" },
-            { Rgba32.Transparent, "255,255,255,0" },
-            { Rgba32.Orange, "orange" },
-            { Rgba32.Lime, "00FF00" },
+            { Color.White, "255,255,255" },
+            { Color.Transparent, "0,0,0,0" },
```
brianpopow: note that i have changed the expected value from transparent from 255, 255, 255,0 to 0,0,0,0 as i think that's the correct value for transparent, see: [transparent](https://www.w3.org/TR/css-color-3/#transparent-def)

README.md
```diff
@@ -157,6 +157,7 @@ Core Team
 - [Dirk Lemstra](https://github.com/dlemstra)
 - [Anton Firsov](https://github.com/antonfirsov)
 - [Scott Williams](https://github.com/tocsoft)
+- [Brian Popow](https://github.com/brianpopow)
```
JimBobSquarePants: Hell yeah 👍

tests/ImageSharp.Web.Tests/Commands/CommandParserTests.cs
JimBobSquarePants: Yep, that's correct. W3C defines transparent as 0,0,0,0 only MS use 255,255,255,0

PR #100, opened 2020-07-07
Converts callbacks to async

wagich:
### Prerequisites

### Description
This PR changes the callbacks on `OnParseCommands`, `OnBeforeSave`, `OnProcessed` and `OnPrepareResponse` on `ImageSharpMiddlewareOptions` to be a `Func<..., Task>` instead of `void` to enable async callbacks. 
See #99 for context.


src/ImageSharp.Web/Middleware/ImageContext.cs
```diff
@@ -125,12 +125,11 @@ public void ComprehendRequestHeaders(DateTimeOffset lastModified, long length)
         /// <param name="statusCode">The status code.</param>
         /// <param name="metaData">The image metadata.</param>
         /// <returns>The <see cref="Task"/>.</returns>
-        public Task SendStatusAsync(int statusCode, in ImageCacheMetadata metaData)
```
wagich, line 128: `async` methods can't have `in` parameters, so I've removed `in` from the signature

src/ImageSharp.Web/Middleware/ResponseConstants.cs
```diff
@@ -26,17 +26,5 @@ internal static class ResponseConstants
...
-        /// An empty completed task.
-        /// </summary>
-        internal static readonly Task CompletedTask = CreateCompletedTask();
-
-        private static Task CreateCompletedTask()
-        {
-            var tcs = new TaskCompletionSource<object>();
-            tcs.SetResult(null);
-            return tcs.Task;
-        }
```
wagich, line 40: This seemed to be only referenced by `ImageContext.SendStatusAsync` which is now async.

JimBobSquarePants:
@sebastienros @deanmarcussen I'd love to have your input on this change if you have the chance as it's a breaking change? I want to ensure we have exhausted all other alternatives before doing so.

See #99 for the context.

deanmarcussen:
Makes sense for the use case @wagich . Kinda cool use as well.

Why not `ValueTask` ? I can't remember what frameworks you are still supporting

From memory when @lahma looked at this area it was quite a hot path.

It should probably become `OnBeforeSaveAsync` etc though as it is breaking, and that makes the breaking more obvious.

wagich:
@deanmarcussen 
I had the [`OpenIdConnectEvents`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.openidconnect.openidconnectevents?view=aspnetcore-3.0) class in mind when making the change, which has a bunch of async callbacks without the suffix.
But using the suffix would open up the possibility of not taking the breaking change at all and having both an async and sync callback – but I don't know if that would be beneficial.

In regards to `ValueTask`: in my (admittedly limited) understanding, `ValueTask` is only useful if my method runs synchronously most of the time _and_ returns a value. Since we're effectively returning nothing, we can just use `Task.CompletedTask` and avoid the allocation. But I'm out of my depth here…

lahma:
With awaits you'll get more state machines and for OC I guess they complete immediately. So on hot paths you would like to check for completion if it's the usual case.

I've done such checks for OC and Fluid, here's Fluid's PR with benchmarks to illustrate the impact: https://github.com/sebastienros/fluid/pull/156

In OC those stacks were deeeep 🙂

JimBobSquarePants:
Ah @lahma sorry I didn't tag you originally, that PR is insightful!

@deanmarcussen Yeah we definitely want to adopt the `**Async` suffix. @wagich I'd say make the breaking change to avoid double listening on hot paths. 

Regarding `ValueTask`, we can certainly support that since the lowest target framework is NET Core 2.1. I have wondered whether we should have been using it in a lot more places actually and am happy to accept any advice you can all offer API wise there. 

I'd love to get some sort of throughput tests in place tbh so we can track potential improvements.

wagich:
I did some rudimentary performance testing, but keep in mind that my first contact with BenachmarkDotNet was about 3 hours ago:
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.100-preview.5.20279.10
  [Host] : .NET Core 3.1.5 (CoreCLR 4.700.20.26901, CoreFX 4.700.20.27001), X64 RyuJIT

Job=InProcess  Toolchain=InProcessEmitToolchain  

```
|                      Method |     Mean |   Error |   StdDev | Ratio | RatioSD |  Gen 0 |  Gen 1 | Gen 2 | Allocated |
|---------------------------- |---------:|--------:|---------:|------:|--------:|-------:|-------:|------:|----------:|
|            &#39;Sync Callbacks&#39; | 320.1 us | 6.34 us | 12.37 us |  1.00 |    0.00 | 4.3945 |      - |     - |  38.75 KB |
|      &#39;Async Task Callbacks&#39; | 329.0 us | 6.54 us |  9.17 us |  1.02 |    0.05 | 4.8828 | 0.4883 |     - |  39.03 KB |
| &#39;Async ValueTask Callbacks&#39; | 322.3 us | 6.33 us | 11.42 us |  1.00 |    0.06 | 4.8828 | 0.4883 |     - |  39.31 KB |

The benchmarking code can be found [in this gist](https://gist.github.com/wagich/b5f0f535ab61a881153d48c1087a7d7e), I've used `imagesharp-logo.png` from the test project.

lahma:
The benchmark is a bit problematic because as far as I can tell, it tests the whole pipeline and with that for example metrics of memory usage originate from the HttpClient.

Real world throughput tests with full pipeline would require separate hardware for server and creating the load.

Ideally we would benchmark the middleware directly.

wagich:
Yeah, I tried that at first – but after needing to mock about 20 different things (mostly asp.net core infrastructure) I honestly gave up. I was thinking this way we'd at least get a comparison to the baseline.
I've since refined the code somewhat and also properly changed-out the 4th callback in `ImageContext`, these are the new results:
``` ini

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
.NET Core SDK=5.0.100-preview.5.20279.10
  [Host] : .NET Core 3.1.5 (CoreCLR 4.700.20.26901, CoreFX 4.700.20.27001), X64 RyuJIT

Server=True  Toolchain=InProcessEmitToolchain  

```
|                      Method |     Mean |   Error |   StdDev |   Median | Ratio | RatioSD |  Gen 0 | Gen 1 | Gen 2 | Allocated |
|---------------------------- |---------:|--------:|---------:|---------:|------:|--------:|-------:|------:|------:|----------:|
|            &#39;Sync Callbacks&#39; | 311.3 us | 6.17 us | 17.41 us | 304.3 us |  1.00 |    0.00 | 4.3945 |     - |     - |  38.74 KB |
|      &#39;Async Task Callbacks&#39; | 318.7 us | 4.75 us |  4.44 us | 316.9 us |  0.97 |    0.05 | 4.8828 |     - |     - |  39.03 KB |
| &#39;Async ValueTask Callbacks&#39; | 333.3 us | 6.43 us |  6.01 us | 334.6 us |  1.02 |    0.05 | 4.8828 |     - |     - |  39.34 KB |

They seem to have quite a bit of run to run variance, which I suspect also is due to including too much pipeline code.
The complete benchmarking code is in https://github.com/wagich/ImageSharp.Web/tree/async-events-benchmark

Does anyone know if we can pool `ValueTask`? Right now there's a lot of `return new ValueTask;` in the callbacks.

JimBobSquarePants:
@wagich The functional equivalent of `Task.CompletedTask` for `ValueTask` is simply `default`.  You could have a static member reflecting that shared across default implementations until it's finally added to the [framework](https://github.com/dotnet/runtime/issues/27960).

@sebastienros  When do we get to play with [Crank](https://www.youtube.com/watch?v=QIDxGbSJhBE&list=PL1rZQsJPBU2St9-Mz1Kaa7rofciyrwWVx)?  it's been a few weeks since the standup and I can't see it in the dotnet project list.

sebastienros:
@JimBobSquarePants crank is still waiting for a final approval from management ... should have been the easiest part but maybe because "summer vacations" or "covid" we are still waiting.

But I don't think it would be hard to create a custom unit test that would run a benchmark, even without crank. Like a functional test, at least to render the result of the run locally, and to test perf changes on the aspnet stack (instead of a micro benchmark). Not as good as having a dedicated machine to do it, and being able to get traces and store results ... but would still be useful.

JimBobSquarePants:
@sebastienros Thanks for getting back so quickly... Yeah, all possible, just a question of time investment.

JimBobSquarePants:
@wagich I haven't forgotten this. Just doing some prep work for the upcoming v1 ImageSharp release first.

wagich:
Regarding the discussion about `Task` vs. `ValueTask`, I was just watching a [GitHub Quick Review](https://youtu.be/pgAFuCF6Kp8?t=1790) where they discussed a similar situation/API and they advocate quite strongly in favor of using `Task` (and therefore returning `Task.CompletedTask`) when there is no return value unless the implementation was doing some specific work using a custom task source.

JimBobSquarePants:
@wagich That's really useful information!

We've shipped ImageSharp v1 now so I can work with you to get this completed and merged in.

deanmarcussen:
@wagich that is super interesting, going to have a watch shortly.

One thing I noticed when I read this pr, just because you changed the invokes of these functions from 
```
this.options.OnBeforeSave?.Invoke(image);
```

```
if (this.options.OnBeforeSaveAsync != null)
{
    await this.options.OnBeforeSaveAsync.Invoke(image);
}
```

That actually these aren't even noops, because they are null checked before hand. 

The sample code defines empty funcs, rather than null, so the null check isn't exactly achieving much but is good :)

But if they were just null, then the task vs valuetask doesn't come into play so much, as for the most part its about the noop.

Just a thought from what I saw as I read the code.

wagich:
@deanmarcussen regarding the null check: in the tests, they are explicitly initialized to null. IMHO it would be simpler to just have the declaration be
```c#
public class ImageSharpMiddlewareOptions {
    public Func<ImageCommandContext, Task> OnParseCommandsAsync { get; set; } => Task.CompletedTask;
    // ... other callbacks
}
```
and drop the null check, but I tried to preserve the existing behavior in the tests.

tocsoft:
Can't drop the null check unless we also added validation to the setter to ensure its never manually set to null.

src/ImageSharp.Web/Middleware/ImageContext.cs
JimBobSquarePants: The method doesn't actually have to be async, we can pass the Task along directly.

JimBobSquarePants:
@wagich I took the liberty of updating your branch and completing this PR. 

I won't merge it until the morning now as I want anyone interested to have a look at the changes first. I opted for `Task` with validation in the options so we could drop the checks in the middleware.

JimBobSquarePants (approved):

tocsoft (approved):
lgtm too

wagich:
Thank you @JimBobSquarePants and every one else involved for getting this in!

PR #106, opened 2020-08-23
Introduce ArrayPoolStream

JimBobSquarePants:
### Prerequisites

### Description
Replaces  `ChunkedMemoryStream` with a higher performance complete `ArrayPoolStream` implementation. I'm deliberately using the `ArrayPool` over `MemoryAllocator` to ensure pooling always happens independently of `Configuration` setup.

The type is required to be public so that provider extensions can take advantage of it.

This improves allocations and performance; for sizes over the pool's upper limit, it tracks `MemoryStream` for perf etc.

|          Method |    Bytes |             Mean |            Error |           StdDev |           Median | Ratio | RatioSD |     Gen 0 |     Gen 1 |     Gen 2 |   Allocated |
|---------------- |--------- |-----------------:|-----------------:|-----------------:|-----------------:|------:|--------:|----------:|----------:|----------:|------------:|
|    MemoryStream |        0 |         48.57 ns |         0.398 ns |         0.373 ns |         48.59 ns |  1.00 |    0.00 |    0.0172 |         - |         - |        72 B |
| ArrayPoolStream |        0 |         57.65 ns |         0.713 ns |         0.595 ns |         57.68 ns |  1.19 |    0.02 |    0.0172 |    0.0001 |         - |        72 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream |      100 |         92.20 ns |         1.218 ns |         1.079 ns |         91.99 ns |  1.00 |    0.00 |    0.0842 |         - |         - |       352 B |
| ArrayPoolStream |      100 |        108.85 ns |         1.035 ns |         0.918 ns |        108.74 ns |  1.18 |    0.02 |    0.0172 |         - |         - |        72 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream |     1000 |        177.21 ns |         5.906 ns |        17.229 ns |        170.94 ns |  1.00 |    0.00 |    0.2620 |         - |         - |      1096 B |
| ArrayPoolStream |     1000 |        163.75 ns |         1.163 ns |         1.088 ns |        163.60 ns |  0.97 |    0.05 |    0.0172 |         - |         - |        72 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream |    10000 |      2,662.05 ns |        36.780 ns |        28.715 ns |      2,651.90 ns |  1.00 |    0.00 |    7.3776 |    0.0038 |         - |     30891 B |
| ArrayPoolStream |    10000 |        949.87 ns |        18.911 ns |        19.420 ns |        952.24 ns |  0.36 |    0.01 |    0.0172 |         - |         - |        72 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream |   100000 |     66,324.20 ns |     1,206.548 ns |     1,069.574 ns |     66,307.68 ns |  1.00 |    0.00 |   41.6260 |   41.6260 |   41.6260 |    260351 B |
| ArrayPoolStream |   100000 |      6,233.35 ns |       194.150 ns |       181.608 ns |      6,166.53 ns |  0.09 |    0.00 |    0.0153 |         - |         - |        72 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream |  1000000 |    928,935.06 ns |     8,751.441 ns |     7,757.925 ns |    929,483.59 ns |  1.00 |    0.00 |  499.0234 |  499.0234 |  499.0234 |   2095596 B |
| ArrayPoolStream |  1000000 |     68,249.88 ns |       702.734 ns |       622.955 ns |     68,039.09 ns |  0.07 |    0.00 |         - |         - |         - |        72 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream |  5000000 |  7,188,608.08 ns |   147,084.294 ns |   201,330.642 ns |  7,115,904.69 ns |  1.00 |    0.00 |  742.1875 |  742.1875 |  742.1875 |  16775638 B |
| ArrayPoolStream |  5000000 |  6,227,567.79 ns |    61,707.314 ns |    51,528.402 ns |  6,204,451.56 ns |  0.86 |    0.03 |  406.2500 |  406.2500 |  406.2500 |  14680278 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream | 10000000 | 14,220,921.25 ns |   241,265.703 ns |   225,680.088 ns | 14,198,512.50 ns |  1.00 |    0.00 | 1484.3750 | 1484.3750 | 1484.3750 |  33552960 B |
| ArrayPoolStream | 10000000 | 13,837,654.90 ns |   273,710.823 ns |   457,309.298 ns | 13,746,661.72 ns |  0.98 |    0.04 |  640.6250 |  640.6250 |  640.6250 |  31457561 B |
|                 |          |                  |                  |                  |                  |       |         |           |           |           |             |
|    MemoryStream | 50000000 | 65,091,715.00 ns |   798,369.233 ns |   746,795.075 ns | 65,071,037.50 ns |  1.00 |    0.00 | 3875.0000 | 3875.0000 | 3875.0000 | 134216304 B |
| ArrayPoolStream | 50000000 | 64,581,138.59 ns | 1,282,853.263 ns | 1,622,398.625 ns | 63,905,000.00 ns |  1.00 |    0.03 | 1875.0000 | 1875.0000 | 1875.0000 | 132120960 B |


antonfirsov:
`ArrayPool.Default` 's size limit is 1MB. Over that limit, the pool will *always* create a new array. If you look at the benchmark results, there is a proof that the two solutions are equivalent over 1MB, the small difference in allocation speed is only because `ArrayPool` utilizes `GC.AllocateUninitializedArray<T>`.

A significant portion of web thumbnail making scenarios will deal with inputs which fell outside of this limit, those will not benefit at all from this solution. (Think of high-quality photo collection sites like carbonmade!)

I'm not owner of ImageSharp.Web, but I strongly discourage proceeding with `ArrayPoolStream`, and rethink the approach. I think `ArrayPoolStream` is a poor experimental implementation implementation that underestimates the complexity of the problem, and has ugly workarounds with the bucket roundup.

[`RecyclableMemoryStream`](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream) is superior in all ways. Well designed, well tested with  [high adaption rates](https://www.nuget.org/packages/Microsoft.IO.RecyclableMemoryStream/) Why not using that instead?

James-Jackson-South:
I think your description of the implementation is a shade harsh. It's a long way from the POC and I've worked hard to make it pass every test (and more) than `MemoryStream` does in the CLR. The stream accepts an `ArrayPool` instance to allow for configuration which we can easily pass through middleware options.

>RecyclableMemoryStream is superior in all ways. Well designed, well tested with high adaption rates. Why not using that instead?

Because in order to use that in it's current form we have to maintain a static `RecyclableMemoryStreamManager` for people to consume. If someone doesn't use it then everything breaks down. 

I could theoretically make some sort of wrapper stream for that though and proxy every method.

JimBobSquarePants:
I think your description of the implementation is a shade harsh. It's a long way from the POC and I've worked hard to make it pass every test (and more) than `MemoryStream` does in the CLR. The stream accepts an `ArrayPool` instance to allow for configuration which we can easily pass through middleware options.

>RecyclableMemoryStream is superior in all ways. Well designed, well tested with high adaption rates. Why not using that instead?

Because in order to use that in it's current form we have to maintain a static `RecyclableMemoryStreamManager` for people to consume. If someone doesn't use it then everything breaks down. 

I could theoretically make some sort of wrapper stream for that though and proxy every method.

JimBobSquarePants:
@antonfirsov Here's a demonstration wrapper class. My concern here is that if, for some reason someone chooses not to use our `RecyclableMemoryStreamManager` instance (not recommended of course) then there are now two large pools sitting around. 

``` c#
/// <summary>
/// An in-memory stream similar to <see cref="MemoryStream"/>, that rents contiguous
/// buffers from a <see cref="RecyclableMemoryStreamManager"/> instance.
/// </summary>
public sealed class PooledMemoryStream : Stream
{
    private static readonly RecyclableMemoryStreamManager DefaultMemoryManager = new RecyclableMemoryStreamManager();
    private readonly RecyclableMemoryStream memoryStream;
    private bool isDisposed;

    /// <summary>
    /// Initializes a new instance of the <see cref="PooledMemoryStream"/> class.
    /// </summary>
    public PooledMemoryStream()
        : this(DefaultMemoryManager)
    {
    }

    /// <summary>
    /// Initializes a new instance of the <see cref="PooledMemoryStream"/> class.
    /// </summary>
    /// <param name="memoryManager">The memory manager to use.</param>
    public PooledMemoryStream(RecyclableMemoryStreamManager memoryManager)
        => this.memoryStream = new RecyclableMemoryStream(memoryManager);

    /// <inheritdoc/>
    public override bool CanRead => this.memoryStream.CanRead;

    /// <inheritdoc/>
    public override bool CanSeek => this.memoryStream.CanSeek;

    /// <inheritdoc/>
    public override bool CanWrite => this.memoryStream.CanWrite;

    /// <inheritdoc/>
    public override bool CanTimeout => this.memoryStream.CanTimeout;

    /// <inheritdoc/>
    public override long Length => this.memoryStream.Length;

    /// <inheritdoc/>
    public override long Position
    {
        get => this.memoryStream.Position;

        set => this.memoryStream.Position = value;
    }

    /// <inheritdoc/>
    public override int ReadTimeout
    {
        get => this.memoryStream.ReadTimeout;
        set => this.memoryStream.ReadTimeout = value;
    }

    /// <inheritdoc/>
    public override int WriteTimeout
    {
        get => this.memoryStream.WriteTimeout;
        set => this.memoryStream.WriteTimeout = value;
    }

    /// <inheritdoc/>
    public override void Flush()
        => this.memoryStream.Flush();

    /// <inheritdoc/>
    public override Task FlushAsync(CancellationToken cancellationToken)
        => this.memoryStream.FlushAsync(cancellationToken);

    /// <inheritdoc/>
    public override int Read(byte[] buffer, int offset, int count)
        => this.memoryStream.Read(buffer, offset, count);

    /// <inheritdoc/>
    public override int Read(Span<byte> buffer) => this.memoryStream.Read(buffer);

    /// <inheritdoc/>
    public override Task<int> ReadAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
        => this.memoryStream.ReadAsync(buffer, offset, count, cancellationToken);

    /// <inheritdoc/>
    public override ValueTask<int> ReadAsync(Memory<byte> buffer, CancellationToken cancellationToken = default)
        => this.memoryStream.ReadAsync(buffer, cancellationToken);

    /// <inheritdoc/>
    public override int ReadByte() => this.memoryStream.ReadByte();

    /// <inheritdoc/>
    public override IAsyncResult BeginRead(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
        => this.memoryStream.BeginRead(buffer, offset, count, callback, state);

    /// <inheritdoc/>
    public override int EndRead(IAsyncResult asyncResult) => this.memoryStream.EndRead(asyncResult);

    /// <inheritdoc/>
    public override long Seek(long offset, SeekOrigin origin)
        => this.memoryStream.Seek(offset, origin);

    /// <inheritdoc/>
    public override void SetLength(long value)
        => this.memoryStream.SetLength(value);

    /// <inheritdoc/>
    public override void Write(byte[] buffer, int offset, int count)
        => this.memoryStream.Write(buffer, offset, count);

    /// <inheritdoc/>
    public override void Write(ReadOnlySpan<byte> buffer) => this.memoryStream.Write(buffer);

    /// <inheritdoc/>
    public override Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken)
        => this.memoryStream.WriteAsync(buffer, offset, count, cancellationToken);

    /// <inheritdoc/>
    public override ValueTask WriteAsync(ReadOnlyMemory<byte> buffer, CancellationToken cancellationToken = default)
        => this.memoryStream.WriteAsync(buffer, cancellationToken);

    /// <inheritdoc/>
    public override void WriteByte(byte value)
        => this.memoryStream.WriteByte(value);

    /// <inheritdoc/>
    public override IAsyncResult BeginWrite(byte[] buffer, int offset, int count, AsyncCallback callback, object state)
        => this.memoryStream.BeginWrite(buffer, offset, count, callback, state);

    /// <inheritdoc/>
    public override void EndWrite(IAsyncResult asyncResult)
        => this.memoryStream.EndWrite(asyncResult);

    /// <inheritdoc/>
    public override void Close() => this.memoryStream.Close();

    /// <inheritdoc/>
    public override void CopyTo(Stream destination, int bufferSize) => this.memoryStream.CopyTo(destination, bufferSize);

    /// <inheritdoc/>
    public override Task CopyToAsync(Stream destination, int bufferSize, CancellationToken cancellationToken)
        => this.memoryStream.CopyToAsync(destination, bufferSize, cancellationToken);

    /// <inheritdoc/>
    public override ValueTask DisposeAsync() => this.memoryStream.DisposeAsync();

    /// <inheritdoc/>
    public override bool Equals(object obj) => this.memoryStream.Equals(obj);

    /// <inheritdoc/>
    public override int GetHashCode() => this.memoryStream.GetHashCode();

    /// <inheritdoc/>
    public override object InitializeLifetimeService()
        => this.memoryStream.InitializeLifetimeService();

    /// <inheritdoc/>
    public override string ToString()
        => this.memoryStream.ToString();

    /// <inheritdoc/>
    protected override void Dispose(bool disposing)
    {
        if (this.isDisposed)
        {
            return;
        }

        this.isDisposed = true;
        this.memoryStream.Dispose();
    }
}
```

antonfirsov:
> My concern here is that if, for some reason someone chooses not to use our RecyclableMemoryStreamManager instance (not recommended of course) then there are now two large pools sitting around.

This thing should be DI-d, so users can replace it in their configuration if they want.

> I think your description of the implementation is a shade harsh. It's a long way from the POC and I've worked hard to make it pass every test (and more) than `MemoryStream` does in the CLR

I'm sorry, I didn't mean to devalue your work adapting it, it's more like a criticism towards the original approach in `Pipelines.Sockets.Unofficial`.  

I'm working on a longer comment explaining my arguments.

JimBobSquarePants:
> This thing should be DI-d, so users can replace it in their configuration if they want.

I should drink more coffee. That's such an obvious solution.

antonfirsov:
So:

There is a general design question that needs to be solved, when addressing the buffer/array pooling problem in the field of image processing, where buffers can grow much larger than in the poolings problems present in the field of networking, and the related protocol parsing problems:

*How to provide a buffer "pool" that works well with both small and **very large** buffers?*

In my understanding the the design of BCL array pools is unfortunately overfocused on the networking/parsing problems, and doesn't really try to solve the problem above. In `ArrayPoolMemoryAllocator` I tried to address this by having two pools around. One for "normal" and one for very large buffers. It's much better than nothing, but I'm still unsure if it's optimal.

One big advantage of `RecyclableMemoryStream` that it seems do be designed [to deal with all this complexity](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream/blob/master/poolcomparison.png). Note: `maximumBufferSize` should be over 2MB in our default manager.

JimBobSquarePants:
@antonfirsov Take a look at #107 when you get a chance please. Much better.

antonfirsov:
@JimBobSquarePants will do in the evening, or tomorrow latest.

PR #107, opened 2020-08-24
Use RecyclableMemoryStream

JimBobSquarePants:
### Prerequisites

A replacement for #106 that uses `RecyclableMemoryStream` for buffer pooling.

Note. Suggestions welcome for default pooling configuration values.


antonfirsov (approved):
Looks good, and we are likely fine with the default setup.

`RecyclableMemoryStream` [docs](https://github.com/microsoft/Microsoft.IO.RecyclableMemoryStream#how-it-works) state:
> Large Pool - Holds large buffers, which are only used when you must have a single, contiguous buffer, such as when you plan to call GetBuffer().

So, if no-one will invoke `stream.GetBuffer()`, all streams are puzzled together from 128 kbyte, small buffers, and the large pools remain untouched.

Note the reasoning behind that buffer size selection:

> RecyclableMemoryStream improves GC performance by ensuring that the larger buffers used for the streams are put into the gen 2 heap and stay there forever. This should cause full collections to happen less frequently. If you pick buffer sizes above 85,000 bytes, then you will ensure these are placed on the large object heap, which is touched even less frequently by the garbage collector.

We need to adapt this size in `ChunkedMemoryStream`!

JimBobSquarePants:
>We need to adapt this size in ChunkedMemoryStream!

Yep!

PR #110, opened 2020-08-27
Simplify middleware workflow.

JimBobSquarePants:
### Prerequisites

### Description
Simplifies the middleware workflow and removes the requirement for stream copying in individual providers.
Adds tests to ensure we handle cached and non-cached scenarios.


src/ImageSharp.Web.Providers.Azure/Resolvers/AzureBlobStorageImageResolver.cs
```diff
@@ -16,22 +15,13 @@ namespace SixLabors.ImageSharp.Web.Resolvers.Azure
     /// </summary>
     public class AzureBlobStorageImageResolver : IImageResolver
```
JimBobSquarePants, line 16: Code coverage is reporting a delta of -12.31% for this file, however coverage is 100%. I've read their docs, I don't get it.

JimBobSquarePants:
I'm gonna go rogue here and merge this. I've got limited availability to work on the library so want to press on while I can.

PR #111, opened 2020-08-29
Update Cache Config To Use TimeSpan

JimBobSquarePants:
### Prerequisites

### Description
Updates configuration to allow caching by any `TimeSpan` configuration. See #104 

This is a necessary breaking change.

I've also reintroduced the samples webapp into the solution.


tocsoft (commented):
I do think I would just make more sense to just have a single option called `MaxCacheTimeout` and it be of type `TimeSpan` instead of the now 3 separate parts it then allows for super fine grained adjustments without have to keep adding additional properties and having to then coherence the multiple files into a `TimeSpan` anyway.

JimBobSquarePants:
Yeah, we'd have to determine a format though that can be parsed invariantly.

tocsoft:
`TimeSpan`s just work for serializing/parsing, if we are loading from `app.config`s anyway.

tocsoft:
https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-timespan-format-strings

The invariant format looks like this `3.17:25:30.5000000` `[d'.']hh':'mm':'ss['.'fffffff]`  so generally HH:MM:SS with optional days prefix and milliseconds postfix

JimBobSquarePants:
> TimeSpans just work for serializing/parsing, if we are loading from app.configs anyway.

Beautiful. I'll make the changes.

.github/workflows/build-and-test.yml
```diff
@@ -74,7 +74,7 @@ jobs:
                   XUNIT_PATH: .\tests\ImageSharp.Web.Tests # Required for xunit
 
             - name: Update Codecov
-              uses: codecov/codecov-action@v1.0.7
+              uses: codecov/codecov-action@v1
```
JimBobSquarePants, line 77: We'll have to do this in all our repos. They've changed things which causes a 400

tocsoft (approved):
much better 👍

PR #112, opened 2020-08-30
Azure Blob Cache and Test refactor.

JimBobSquarePants:
### Prerequisites

### Description
This has grown arms and legs. Started off as a simple Azure Blob Storage cache implementation turned into a realization that we're loading up the cache stream for 304s and that all our integration tests are super janky once the perf jumped. (We were generating multiple server instances that were hitting the same cache file).

No breaking changes thankfully, old cache files containing no content length will simply be reprocessed and cached.


src/ImageSharp.Web.Providers.Azure/Caching/AzureBlobStorageCache.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Initializes a new instance of the <see cref="AzureBlobStorageCache"/> class.
+        /// </summary>
+        /// <param name="cacheOptions">The cache options.</param>
+        public AzureBlobStorageCache(IOptions<AzureBlobStorageCacheOptions> cacheOptions)
+        {
+            Guard.NotNull(cacheOptions, nameof(cacheOptions));
+            AzureBlobStorageCacheOptions options = cacheOptions.Value;
+
+            this.container = new BlobContainerClient(options.ConnectionString, options.ContainerName);
+            this.container.CreateIfNotExistsAsync(PublicAccessType.None);
```
tocsoft, line 31: should we really be creating this if it doesn't exists? or should we require application to do this during setup, I see this the same as EF etc I wouldn't expect my ORM to create the database for me unless I configured it so. We should however error if it doesn't exist.

Also shouldn't really be doing potentially expensive operations inside the constructor. There should probably be a `private BlobContainerClient GetClientAsync()` that is called to get the client or set it up if it hasn't yet been instantiated.

tocsoft (commented):
Generally looking good, only thing I have a question mark over is the behaviour in `AzureBlobStorageCache ` and generally the way the blob container is managed.

src/ImageSharp.Web.Providers.Azure/Caching/AzureBlobStorageCache.cs
JimBobSquarePants: We could make it optional I suppose. This changes the behavior compared to the physical cache though and our own default behavior of creating directories when we save to a destination path. 

That constructor is only called once, the cache is registered as a singleton.

tocsoft: I feel file system is different you aren't dealing with potentially expensive operations on a remote server, you are dealing with local resources and can assume that is fast.

I think my main issue is that we really shouldn't be being called async/remote APIs during construction.

Even having a static async factory method would be fine its just that network call in the middle of the constructor that bothers me the most.

JimBobSquarePants: I'll move it to a factory then attached to the cache and allow passing the view enum as well then.

tocsoft (approved):

PR #117, opened 2020-09-03
Make converters injectable and respect culture

JimBobSquarePants:
### Prerequisites

### Description
Apologies in advance, this touches a lot of files. Coverage is good though.

This is one of those breaking changes we unfortunately have to implement before V1

The current `CommandParser` was not injectable and also did not respect the current culture. Respecting culture is a required feature for using in many CMS implementations.

By default the system will use `InvariantCulture` so there are no surprises for people upgrading but there is a switch in the middleware options `UseInvariantParsingCulture` which allows using the `CultureInfo.CurrentCulture`.

Converters and Processors both required  method signature changes to allow the update.

I also found and fixed two issues
1. We were disposing of the output stream outside of the write lock
2. We were not returning after a head request.

- [x] Refactor converters, parser, and processors to allow current culture
- [x] Update tests 
- [x] Add additional tests


src/ImageSharp.Web/Commands/CommandParser.cs
```diff
@@ -61,99 +37,38 @@ private CommandParser()
...
-
-            this.AddConverter(typeof(short[]), typeof(ArrayConverter<short>));
-            this.AddConverter(typeof(ushort[]), typeof(ArrayConverter<ushort>));
+            DebugGuard.NotNull(type, nameof(type));
+            DebugGuard.NotNull(culture, nameof(culture));
 
-            this.AddConverter(typeof(int[]), typeof(ArrayConverter<int>));
-            this.AddConverter(typeof(uint[]), typeof(ArrayConverter<uint>));
+            // This allows us to reuse the same converter for infinite enum types.
+            Type matchType = type.IsEnum ? typeof(Enum) : type;
```
tocsoft, line 58: This does mean that users can't register custom Converters for specific enums however... should this be a fall back instead of a force change? i.e. looks for specific converter first, then use the generic Enum converter if one can't be found.

tocsoft (commented):
just the one comment on Enum handling but all the rest is looking good to me.

Only other thing that I have a small issue (mostly an aesthetics thing and not something I would explicitly want changing unless you wanted to) is the DefaultServicesAreRegistered tests... that to me is screaming out to be a theory test taking in the Service and implementation types as arguments.

src/ImageSharp.Web/Commands/CommandParser.cs
JimBobSquarePants: Yep, wasn’t something I considered but can easily workaround. Will update 👍

JimBobSquarePants:
@tocsoft I updated the parser to allow custom enum types also fixed some boxing allocations caused by the parser/converter API

Agree re the test but left it ugly as-is for now as I didn't want to end up potentially fighting Xunit skipping theory data values because it thinks they're all the same.  Time is precious! 😄

src/ImageSharp.Web/Commands/CommandParser.cs
JimBobSquarePants: Done

tocsoft (approved):

PR #121, opened 2020-09-11
WIP Faster Cache

JimBobSquarePants:
### Prerequisites

### Description
Introduces a time aware least recently used cache to store results of cache lookups. This removes most of the IO required to check whether an image is in the cache for subsequent requests for the same URL.

I'm extremely concerned about the way our AsyncKeyLock works now though. Cache lookups are now so fast we're getting misses as the next request for the same URL is attempting to read the cache. This leads to multiple writes which are safe but inefficient.


JimBobSquarePants:
You can actually see the negative effect in the Doorman coverage here as all our small number of requests are entering the read lock before the write lock is established.

https://codecov.io/gh/SixLabors/ImageSharp.Web/compare/aa1b1928c675cf02e86c3a6284b9d6b22d1880e5...620f818383811c4a49075d7317e1fddd1bda3089/src/src/ImageSharp.Web/Caching/Doorman.cs

It seems to me we need a single lock which is upgradeable. Any and all advice is most welcome.

deanmarcussen:
Have you considered trying a lock free (ish because of the Lazy) implementation?

Something like
`private readonly ConcurrentDictionary<string, Lazy<Task>> Workers = new ConcurrentDictionary<string, Lazy<Task>>(StringComparer.OrdinalIgnoreCase);`

JimBobSquarePants:
@deanmarcussen I don't quite follow I'm afraid.

deanmarcussen:
> @deanmarcussen I don't quite follow I'm afraid.

Sorry not well explained - late on a Friday.

Easier to show than explain. Here's a working branch for ideas, not stress tested, or perf tested.

The idea is to maintain a thread safe dictionary of lazy tasks, rather than use a lock.

Ignore if you don't like

https://github.com/ThisNetWorks/ImageSharp.Web/tree/lock-free

JimBobSquarePants:
@deanmarcussen More me than you mate, I'm not firing on all cylinders. 

Thanks for having a look! I'd love to go lock free actually!

 Might be best to base any changes on this branch though as I've changed the cache lookup method to speed it up.

deanmarcussen:
> Might be best to base any changes on this branch though as I've changed the cache lookup method to speed it up. Though if I’m reading your code right your changes would negate the need for an lru cache.

@JimBobSquarePants Trying to get time to look at this properly today.

From pulling this branch it looks like the LRU cache is using similar concepts to what I was proposing for the AsyncLock, but to achieve a different result.

So to clarify, is the LRU is to provide cached access to source metadata and source resolver - not the actual stream but just the resolver?

I think you would still want threadsafe locklike behaviour read around the cache, and write safe when not in the cache?

I'll try and update based of this branch later today

JimBobSquarePants:
Thanks @deanmarcussen 

>So to clarify, is the LRU is to provide cached access to source metadata and source resolver - not the actual stream but just the resolver?

They cache both resolvers and the metadata returned. Reading values from the cache is actually fairly IO intensive and by caching the result in an expiring LRU cache I can cut that all out. 

>I think you would still want threadsafe locklike behaviour read around the cache, and write safe when not in the cache?

Yeah, I reviewed and still think we need that. The current read/write lock is allowing for too many queued writes though. 

Basically if I have 10 requests sent at the same time the first 7 end up processing the file and the last 3 (plus any future requests) use the cached result. This is in extreme (artificial) circumstances of course so I might be worrying about nothing but I still don't like it.

deanmarcussen:
New branch @JimBobSquarePants https://github.com/ThisNetWorks/ImageSharp.Web/tree/dm/lock-free-cache

What's the best way to test this? Draft PR or will you run it.

I see options for improving it - the Lazy is useful on the Writer, to prevent multiple writes, but may not be necesary on the Reader.

But worth seeing if it has anything to gain over your existing `AsyncKeyLock` first before looking at further optimization.

> Basically if I have 10 requests sent at the same time the first 7 end up processing the file and the last 3

I'm not sure where this race is, and if what I've done achieves anything about it
Because there's still a race/gap between a ReadWorker, and creating a WriteWorker, which might be what you're seeing?

But with this, we should see only one write occuring at least I hope so.

They could potentially merge into one set of Workers, which produce a func for reading (then cached), and perform the write if not read?

Still getting my head around the LRU cache ;)

JimBobSquarePants:
Thanks @deanmarcussen I'll have a look tonight. 

This test is useful for tracking reads and writes. I just debug it and add some break points. Comment out the AzureTestImage theory data if you want to just test the physical provider. 

https://github.com/ThisNetWorks/ImageSharp.Web/blob/0efc559c63f2a9f8811f689482cfbc44eb40797b/tests/ImageSharp.Web.Tests/Processing/PhysicalFileSystemCacheServerTests.cs#L108

sebastienros:
If you need me to run a benchmark let me know which branch

JimBobSquarePants:
@deanmarcussen Just pulled down your branch and all tests are passing. In addition I also saw the correct number of writes whic was awesome.

 Cranked up the numbers and everything still looks good. Curious to see how it profiles.

deanmarcussen:
Whew! And cool.

I need to tweak something tomorrow I think. The write func should probably check the read dictionary for operations and await it as well.

Let's see if we can bench it after :)

deanmarcussen:
@JimBobSquarePants Pushed the tweak I wanted to.

The test you pointed out has been quietly running this morning, with some tweaks to process 10+ images of a larger size (10MB to a width of 200px), and either a new image (cache busted) or the same image from the cache (every odd number task).

I can't fault it.

In terms of performance this is not a great way to bench it, but I applied the same test run to this branch here four or five times and the test run averages around 43s.

On my branch it drops to around 15s per test run!

MacOS i7 Quad Core 24GB

JimBobSquarePants:
@deanmarcussen Can you open a PR targeting this branch then? 

I'll merge it then @sebastienros will be able to profile this branch and tell us what the numbers are.

JimBobSquarePants:
@sebastienros This branch `js/faster-cache` is ready to benchmark at your leisure 🙂

sebastienros:
🎉 

It's faster than static files, at least on a cached file. Went from 2K to 60K, with super low latency too.
There are some error, around 0.1% of the requests, but I don't know where they come from yet. Not more or less than before.

JimBobSquarePants:
>It's faster than static files, at least on a cached file. Went from 2K to 60K, with super low latency too.

Holy Moly! @deanmarcussen I owe you any beverage of your choice!

@sebastienros 0.1% errors is fine by me just now. We can diagnose at another time.

deanmarcussen:
> > It's faster than static files, at least on a cached file. Went from 2K to 60K, with super low latency too.
> 
> Holy Moly! @deanmarcussen I owe you any beverage of your choice!

Mind blown - don't know what I was expecting, but nothing like that. Awesome :)

sebastienros:
Let us know when this is on nuget so we can test it in OC.

JimBobSquarePants:
It's on MyGet just now. I'll push to NuGet in the next day or two once I've double checked a few other things.

https://www.myget.org/feed/sixlabors/package/nuget/SixLabors.ImageSharp.Web/1.0.0-rc.3.16

PR #125, opened 2020-09-17
Allow the ability to set the quality of jpegs.

JimBobSquarePants:
### Prerequisites

### Description
Introduces the ability to set the quality of jpegs per request. Commands are sent in the following manner.

See Discussion #123 for reference.

```
{PATH_TO_YOUR_IMAGE}?quality=50
{PATH_TO_YOUR_IMAGE}?format=jpg&quality=50
```

As we introduce image other image formats that support variable quality we can create format specific quality parsers that will not conflict with this one allowing the sharing of the same command key.

image

### DIFF
--- a/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
@@ -67,7 +67,8 @@ namespace SixLabors.ImageSharp.Web.DependencyInjection
 
             builder.AddProcessor<ResizeWebProcessor>()
                    .AddProcessor<FormatWebProcessor>()
-                   .AddProcessor<BackgroundColorWebProcessor>();
+                   .AddProcessor<BackgroundColorWebProcessor>()
+                   .AddProcessor<JpegQualityWebProcessor>();
 
             builder.AddConverter<IntegralNumberConverter<sbyte>>();
             builder.AddConverter<IntegralNumberConverter<byte>>();
--- a/src/ImageSharp.Web/Processors/JpegQualityWebProcessor.cs
+++ b/src/ImageSharp.Web/Processors/JpegQualityWebProcessor.cs
@@ -1,0 +1,49 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.Collections.Generic;
+using System.Globalization;
+using Microsoft.Extensions.Logging;
+using SixLabors.ImageSharp.Formats.Jpeg;
+using SixLabors.ImageSharp.Web.Commands;
+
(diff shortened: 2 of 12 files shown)


src/ImageSharp.Web/Processors/JpegQualityWebProcessor.cs
```diff
@@ -0,0 +1,49 @@
...
+            ILogger logger,
+            IDictionary<string, string> commands,
+            CommandParser parser,
+            CultureInfo culture)
+        {
+            if (commands.ContainsKey(Quality) && image.Format is JpegFormat)
+            {
+                // The encoder clamps any values so no validation is required.
+                int quality = parser.ParseValue<int>(commands.GetValueOrDefault(Quality), culture);
+                image.Encoder = new JpegEncoder() { Quality = quality };
```
tocsoft, line 43: should we be copying all other settings from the version of the encoder already on the `FormattedImage`? in case the user has custom configuration on the version inside the `Configuration` already and only wants the quality changing.

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -67,7 +67,8 @@ public static IImageSharpBuilder AddImageSharp(this IServiceCollection services)
 
             builder.AddProcessor<ResizeWebProcessor>()
                    .AddProcessor<FormatWebProcessor>()
-                   .AddProcessor<BackgroundColorWebProcessor>();
+                   .AddProcessor<BackgroundColorWebProcessor>()
+                   .AddProcessor<JpegQualityWebProcessor>();
```
tocsoft, line 71: Should this really be on by default or is it slightly more niche and thus opt in?

tocsoft (commented):
just a couple of questions/comments, but generally happy with as it unlock the scenario where a user can now switch out the encoder based on any custom logic they might want which is great.

src/ImageSharp.Web/Processors/JpegQualityWebProcessor.cs
JimBobSquarePants: I was wondering that myself... There's only subsampling so I can do that.

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
JimBobSquarePants: I think people will expect it. I know it's commonly used in the old ImageProcessor libraries.

tocsoft: sounds good then.

tocsoft (approved):

PR #127, opened 2020-11-04
Add an option to enable setting a different CacheRoot

stefannikolei:
### Prerequisites

### Description
Adding a new Option for the cache implementation. Now the CacheRoot can be set to an different folder than the webroot

### DIFF
--- a/src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
+++ b/src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
@@ -72,7 +72,7 @@ namespace SixLabors.ImageSharp.Web.Caching
 
             // Allow configuration of the cache without having to register everything.
             this.cacheOptions = cacheOptions != null ? cacheOptions.Value : new PhysicalFileSystemCacheOptions();
-            this.cacheRootPath = GetCacheRoot(this.cacheOptions, environment.WebRootPath);
+            this.cacheRootPath = GetCacheRoot(this.cacheOptions, environment.WebRootPath, environment.ContentRootPath);
             if (!Directory.Exists(this.cacheRootPath))
             {
                 Directory.CreateDirectory(this.cacheRootPath);
@@ -89,13 +89,17 @@ namespace SixLabors.ImageSharp.Web.Caching
         /// </summary>
         /// <param name="cacheOptions">the cache options.</param>
         /// <param name="webRootPath">the webRootPath.</param>
+        /// <param name="contentRootPath">the contentRootPath.</param>
         /// <returns>root path.</returns>
-        internal static string GetCacheRoot(PhysicalFileSystemCacheOptions cacheOptions, string webRootPath)
+        internal static string GetCacheRoot(in PhysicalFileSystemCacheOptions cacheOptions, in string webRootPath, in string contentRootPath)
         {
             var cacheRoot = string.IsNullOrEmpty(cacheOptions.CacheRoot)
                 ? webRootPath
                 : cacheOptions.CacheRoot;
-            return Path.Combine(cacheRoot, cacheOptions.CacheFolder);
+
+            return Path.IsPathFullyQualified(cacheRoot)
+                ? Path.Combine(cacheRoot, cacheOptions.CacheFolder)
+                : Path.GetFullPath(Path.Combine(cacheRoot, cacheOptions.CacheFolder), contentRootPath);
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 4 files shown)


src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
```diff
@@ -72,7 +72,10 @@ public class PhysicalFileSystemCache : IImageCache
 
             // Allow configuration of the cache without having to register everything.
             this.cacheOptions = cacheOptions != null ? cacheOptions.Value : new PhysicalFileSystemCacheOptions();
-            this.cacheRootPath = Path.Combine(environment.WebRootPath, this.cacheOptions.CacheFolder);
+            var cacheRoot = string.IsNullOrEmpty(this.cacheOptions.CacheRoot)
```
JimBobSquarePants, line 75: Nice and none breaking 👍 I wonder if we can tweak it to allow both relative and absolute paths?

We'll need to add some unit tests also to ensure coverage.

stefannikolei: And relativ to which path? To the WebRoot? Any suggestions how to accomplish that?

I can look into tests.

stefannikolei: I would have added some tests. But creating a unit test for this is not so straight forward. I could only imagine to use a package like system.io.abstractions. But that will probably not help for the PhysicalFileProvider..

stefannikolei: I see only one way. We need to get rid of the instantiation of PhysicalFileProvider in the ctor. 

I would suggest to create a new interface for the FileAccess and a implementation which depends on the PhysicalFileProvider. So we can go and test the logic of setting the root of the Cache.

The question for the answer for the relative path is relatively easy 😄 ...

When you have a look at the docs you see that relative paths are not possible
https://docs.microsoft.com/de-de/aspnet/core/fundamentals/file-providers?view=aspnetcore-3.1#physicalfileprovider


So i am waiting for your suggestions how to proceed

JimBobSquarePants: I think it's possible with the current setup. 

You can use `Path.IsPathFullyQualified` to check if the value is absolute. If it is, take it, if not use `Path.GetFullPath` to combine the relative path with the environment `ContentRootPath` absolute path

stefannikolei: i have added some test. SHould i look into the FullyQualified path in this MR or should we tackle this in a different one?

JimBobSquarePants: Let's do it in one. There should be very little to add and we can update your new tests (nice!) to use inline theory data and test both scenarios.

stefannikolei: ok just for clarification. When specifying e.g. "./testfolder/xyz/" as CacheFolder in the options. You want it to be relative to webroot?

JimBobSquarePants: I'd say relative to the application.

JimBobSquarePants (changes requested):
If you drop the `in` modifiers, this looks great. Thanks!

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
```diff
@@ -89,13 +89,17 @@ public class PhysicalFileSystemCache : IImageCache
         /// </summary>
         /// <param name="cacheOptions">the cache options.</param>
         /// <param name="webRootPath">the webRootPath.</param>
+        /// <param name="contentRootPath">the contentRootPath.</param>
         /// <returns>root path.</returns>
-        internal static string GetCacheRoot(PhysicalFileSystemCacheOptions cacheOptions, string webRootPath)
+        internal static string GetCacheRoot(in PhysicalFileSystemCacheOptions cacheOptions, in string webRootPath, in string contentRootPath)
```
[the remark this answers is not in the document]
stefannikolei: Yeah i can drop them. But it is not specifically for structs. Using it just tells the compiler that it is readonly in this scope. That's how i understood it ;) But i remove them ;)

JimBobSquarePants (approved):
Lovely. Thanks!

I'm planning an official release in a few days but you will be able to grab the build from MyGet in the interim once merged.

Issue #140, opened 2021-02-26
Under heavy load the LRUCache can fail to retrieve image metadata


deanmarcussen:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Web
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

Under heavy load the LRUCache may return what seems to be an empty `ImageCacheMetadata` (when I say empty, it is not null, but has no apparent values).

This results in a dictionary error when trying to read the format utilities in `PhysicalFileSystemCacheResolver`
```
               string path = Path.ChangeExtension(
                this.metaFileInfo.PhysicalPath,
                this.formatUtilities.GetExtensionFromContentType(this.metadata.ContentType));
```

### Steps to Reproduce

```
deanmarcussen@dean ImageSharp.Web % ~/Scripts/bombardier -c 100 -d 5s -k "http://localhost:2646/imagesharp-logo.png?width=300"
Bombarding http://localhost:2646/imagesharp-logo.png?width=300 for 5s using 100 connection(s)
[=============================================================================================================================================================================] 5s
Done!
Statistics        Avg      Stdev        Max
  Reqs/sec      3314.89    1101.36    5268.09
  Latency       30.18ms    22.94ms   468.19ms
  HTTP codes:
    1xx - 0, 2xx - 16654, 3xx - 0, 4xx - 0, 5xx - 7
    others - 0
  Throughput:    18.61MB/s
```

On this run 7 500 errors, corresponding to this exception
```
2021-02-26 08:09:33.2911|1|ERROR|Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware|An unhandled exception has occurred while executing the request. System.ArgumentNullException: Value cannot be null. (Parameter 'key')
   at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
[...]

PR #141, opened 2021-02-26
Fix missing cache lookups

deanmarcussen:
### Prerequisites

### Description

Fixes https://github.com/SixLabors/ImageSharp.Web/issues/140

Simple option. 

Merge them into one cache, and treat them as tuples.

Other option is to cache it locally inside `IImageCacheResolver` so that subsequent calls retrieve the same instance.

But I think this involves less changes.

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -418,7 +418,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
                                         {
                                             metadata = await resolver.GetMetaDataAsync();
                                         }
-                                        
+
                                         return (resolver, metadata);
                                     });
 
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -410,7 +410,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
                             (IImageCacheResolver, ImageCacheMetadata) cachedImageResolver = await
                                 CacheResolverLru.GetOrAddAsync(
                                     key,
-                                    async k => 
+                                    async k =>
                                     {
                                         var resolver = await this.cache.GetAsync(k);
                                         ImageCacheMetadata metadata = default;
@@ -418,6 +418,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
                                         {
                                             metadata = await resolver.GetMetaDataAsync();
                                         }
+                                        
                                         return (resolver, metadata);
                                     });
 
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -50,8 +50,8 @@ namespace SixLabors.ImageSharp.Web.Middleware
         /// <summary>
         /// Used to temporarily store cache resolver reads to reduce the overhead of cache lookups.
         /// </summary>
-        private static readonly ConcurrentTLruCache<string, (IImageCacheResolver, ImageCacheMetadata)> CacheResolverLru
-            = new ConcurrentTLruCache<string, (IImageCacheResolver, ImageCacheMetadata)>(1024, TimeSpan.FromMinutes(5));
+        private static readonly ConcurrentTLruCache<string, ValueTuple<IImageCacheResolver, ImageCacheMetadata>> CacheResolverLru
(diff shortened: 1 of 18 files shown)


src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -50,14 +50,8 @@ public class ImageSharpMiddleware
...
         /// </summary>
-        private static readonly ConcurrentTLruCache<string, IImageCacheResolver> CacheResolverLru
-            = new ConcurrentTLruCache<string, IImageCacheResolver>(1024, TimeSpan.FromMinutes(5));
-
-        /// <summary>
-        /// Used to temporarily store cache metadata reads to reduce the overhead of cache lookups.
-        /// </summary>
-        private static readonly ConcurrentTLruCache<string, ImageCacheMetadata> CacheMetadataLru
-            = new ConcurrentTLruCache<string, ImageCacheMetadata>(1024, TimeSpan.FromMinutes(5));
+        private static readonly ConcurrentTLruCache<string, (IImageCacheResolver, ImageCacheMetadata)> CacheResolverLru
```
tocsoft, line 53: I would se you either need to name these tuple values or create a custom struct that encapsulates these two fields.

Reason is it will remove the horrid use of `foo.Item1` & `foo.Item2` field access. with well named fields the follow on code would be significantly more readable.

ie in some code you have later you have `cachedImageResolver.Item2.SourceLastWriteTimeUtc` when reading in isolation there is nothing to hint at what `Item2` means in that line

JimBobSquarePants:
Code Coverage drop is caused by a binding redirect to a buggy version of Coverlet. Updating to the latest submodule and building the solution should fix this

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
JimBobSquarePants: Agreed.

deanmarcussen: Done. 

Hmm, updated the submodule as suggested and the code coverage complaints have skyrocketed :(

JimBobSquarePants:
@deanmarcussen Lemme pull down your branch. I think there mhave been a breaking change in the submodule paths causing the stylecop rule changes

JimBobSquarePants:
@deanmarcussen Can you give me push access to the PR. I don't seem to have them. 

https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

deanmarcussen:
> @deanmarcussen Can you give me push access to the PR. I don't seem to have them.

Invite sent. 

Sorry, it's really annoying github have changed the security settings for organisations, and anything forked to my org doesn't give access to edit pull requests by default.
I need to move all my forks to my personal account.

JimBobSquarePants:
No worries mate. Looks like the xunit upgrade has busted `CommandParserCanConvert<T>(T expected, string param, CultureInfo culture)` unit tests. I'm getting ambiguous match exceptions. 

xunit test thyself!

JimBobSquarePants (approved):
LGTM.

sebastienros:
@deanmarcussen perf numbers before/after ?

sebastienros:
When I sent my previous comment, I was afraid that stylecop would refuse it because it has a slash in it.

deanmarcussen:
Might have won a little bit from it as well.
`http://localhost:2646/imagesharp-logo.png?width=300`

before:
```
Statistics        Avg      Stdev        Max
  Reqs/sec     40457.76    9068.64   54989.01
  Latency        2.47ms     1.90ms   129.56ms
  HTTP codes:
    1xx - 0, 2xx - 805425, 3xx - 0, 4xx - 0, 5xx - 3552
    others - 0
  Throughput:   191.02MB/s
```

after:

```
Statistics        Avg      Stdev        Max
  Reqs/sec     42820.66    6498.54   55995.16
  Latency        2.33ms   689.17us   221.96ms
  HTTP codes:
    1xx - 0, 2xx - 856105, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:   203.31MB/s
```

sebastienros:
How can we measure the max speed of a disk. Wondering if the limit here is how fast it can read the file. Totally unrelated to the PR.

sebastienros:
Apparently on windows you can do:

```
winsat disk -drive c

> Disk  Random 16.0 Read                       789.24 MB/s          8.6
> Disk  Sequential 64.0 Read                   1433.71 MB/s          8.7
> Disk  Sequential 64.0 Write                  646.15 MB/s          8.2
```

JimBobSquarePants:
I'm gonna merge this and also produce a stable release for NuGet to make it easier to consume. Thanks for the assist!

PR #147, opened 2021-03-23
Resolve 145: add option to ignore URL host in cache key

yringler:
### Prerequisites

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -246,7 +246,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
             IDictionary<string, string> commands)
         {
             // Create a cache key based on all the components of the requested url
-            string uri = GetUri(context, commands);
+            string uri = GetUri(context, commands, this.options.IgnoreHost);
             string key = this.cacheHash.Create(uri, this.options.CachedNameLength);
 
             // Check the cache, if present, not out of date and not requiring and update
@@ -492,9 +492,9 @@ namespace SixLabors.ImageSharp.Web.Middleware
             }
         }
 
-        private static string GetUri(HttpContext context, IDictionary<string, string> commands)
+        private static string GetUri(HttpContext context, IDictionary<string, string> commands, bool ignoreHost)
         {
-            var sb = new StringBuilder(context.Request.Host.ToString());
+            var sb = new StringBuilder(ignoreHost ? null : context.Request.Host.ToString());
 
             string pathBase = context.Request.PathBase.ToString();
             if (!string.IsNullOrWhiteSpace(pathBase))
--- a/samples/ImageSharp.Web.Sample/Startup.cs
+++ b/samples/ImageSharp.Web.Sample/Startup.cs
@@ -151,6 +151,11 @@ namespace SixLabors.ImageSharp.Web.Sample
                 app.UseDeveloperExceptionPage();
             }
 
+            app.Use(async (context, next) =>
+            {
+                context.Request.Host = new Microsoft.AspNetCore.Http.HostString("jjjjjj.org");
+                await next();
+            });
             app.UseDefaultFiles();
             app.UseImageSharp();
             app.UseStaticFiles();
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddlewareOptions.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddlewareOptions.cs
@@ -91,6 +91,11 @@ namespace SixLabors.ImageSharp.Web.Middleware
         public uint CachedNameLength { get; set; } = 12;
 
         /// <summary>
+        /// Gets or sets a value indicating whether to use the host portion of the request URI when generating the key used for the image cache.
+        /// </summary>
+        public bool IgnoreHost { get; set; }
+
+        /// <summary>
         /// Gets or sets the additional command parsing method that can be used to used to augment commands.
         /// This is called once the commands have been gathered and before an <see cref="IImageProvider"/> has been assigned.
         /// </summary>
--- a/tests/ImageSharp.Web.Tests/Processing/AzureBlobStorageCacheServerTests.cs
+++ b/tests/ImageSharp.Web.Tests/Processing/AzureBlobStorageCacheServerTests.cs
@@ -27,78 +27,7 @@ namespace SixLabors.ImageSharp.Web.Tests.Processing
         [Theory]
         [InlineData(TestConstants.PhysicalTestImage)]
         [InlineData(TestConstants.AzureTestImage)]
-        public async Task CanProcessAndResolveImageAsync(string url)
-        {
-            string ext = Path.GetExtension(url);
-            IImageFormat format = Configuration.Default.ImageFormatsManager.FindFormatByFileExtension(ext);
-
-            // First response
-            HttpResponseMessage response = await this.HttpClient.GetAsync(url + Command);
-
-            Assert.NotNull(response);
-            Assert.True(response.IsSuccessStatusCode);
-            Assert.True(response.Content.Headers.ContentLength > 0);
-            Assert.Equal(format.DefaultMimeType, response.Content.Headers.ContentType.MediaType);
-
-            (Image Image, IImageFormat Format) actual = await Image.LoadWithFormatAsync(await response.Content.ReadAsStreamAsync());
-            using Image image = actual.Image;
-
-            Assert.Equal(Width, image.Width);
-            Assert.Equal(format, actual.Format);
-
-            response.Dispose();
-
-            // Cached Response
-            response = await this.HttpClient.GetAsync(url + Command);
-
-            Assert.NotNull(response);
-            Assert.True(response.IsSuccessStatusCode);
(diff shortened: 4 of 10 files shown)


yringler:
I wasn't able to find tests for the ImageSharpMiddleware, so I didn't add a test.
I did modify the sample Startup.cs (just locally) to verify manually that
1. By default, different hosts have a different cache folder (I added a `.Use()` to set the Request.Host)
2. When the option is set, different hosts still use the same cache folder.

yringler:
I see that code coverage is down, but it's tricky for me to find the tests responsible because I can't run all the tests.
`PhysicalFileSystemCacheServerTests` (And AzureBlob) need a localhost server to be running, but there doesn't seem to be a test server?

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -492,9 +492,9 @@ private ValueTask StreamDisposeAsync(Stream stream)
             }
         }
 
-        private static string GetUri(HttpContext context, IDictionary<string, string> commands)
+        private static string GetUri(HttpContext context, IDictionary<string, string> commands, bool ignoreHost)
```
sebastienros, line 495: Quick note that this code can be much improved. We know the size of all strings, so we can allocate the right buffer size instead of growing it in the SB.
Also using the Format("{0}/") is too much for a simple concatenation.
Only `QueryString.Create(commands)` needs to be done before to know what will be the size.
Should I file an issue or not worth it?

JimBobSquarePants: Yep, that's a valid point. Please file an issue. I'm up for any optimizations we can through at the library.

yringler:
I'll be out for the next week and a bit (Passover), so I won't get to the unit tests before that.
If you choose to take care of the tests, I'd be much obliged!

deanmarcussen:
Can I just offer a thought as well. Not sure if this would be best suited via a setting, or by putting this method into an interface, and making it a replaceable component to meet any variety of use cases.

In any kind of multi tenant scenario, where the host name actually means a different tenant / image source it's a little dangerous, easy to set a setting and have two tenants serving the same cached image, when the image source is actually different.

I'm 50/50 on the fence about it though, so will leave design decisions up to @JimBobSquarePants

JimBobSquarePants:
>In any kind of multi tenant scenario, where the host name actually means a different tenant / image source it's a little dangerous, easy to set a setting and have two tenants serving the same cached image, when the image source is actually different.

I get your point here @deanmarcussen a setting is a little too easy to abuse.

I'll have a look at rewriting the PR to introduce two implementations of a as yet unnamed interface that can be injected to allow this. It's a fair amount of work but worth the effort.

yringler:
I added tests for the new option...
But this may need some touching up by a maintainer to get it landed.

> I'll have a look at rewriting the PR to introduce two implementations of a as yet unnamed interface that can be injected to allow this. It's a fair amount of work but worth the effort.

Especially if this has to happen!

JimBobSquarePants:
It’ll be a few days til I can look at this. Sick at the moment with oral shingles.

ronaldbarendse:
This issue kind of surprised me, as it's very common to have your site served from both the `www.` and root domain (and use `canonical` HTML tags to point to the preferred one, instead of redirecting). This will therefore also result in duplicate cached images, which I confirmed by testing on a local site myself (just swapping `localhost` with your local PC name or IP)...

Wouldn't it make a lot more sense to only use the path, as the default setup/middleware configuration doesn't support a multi-tenant setup anyway? If you want to add that support, you already need to change the default configuration, right?

Also note that the [`Request.Host` property might contain the port](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest.host?view=net-6.0#remarks), so it's a bit ambiguous what the new `ignoreHost` parameter might actually ignore.

-------------

The easy/quickest fix would be to change `ICacheHash.Create(string value, uint length)` to `ICacheHash.Create(Uri uri, uint length)` and let the `ImageSharpMiddleware` generate the absolute URI based on the HTTP context and parsed/known commands. However, this assumes both the `IRequestParser` and `IImageProvider` only use the URI to parse/provide the returned value. If you want to return a different image based on whether e.g. a user is logged in or not, this wouldn't work!

So I would suggest adding a new interface that generates the cache key, so this can be separated from the existing `ICacheHash` implementation/responsibility:

```c#
using Microsoft.AspNetCore.Http;
using SixLabors.ImageSharp.Web.Commands;

namespace SixLabors.ImageSharp.Web.Caching
{
    public interface ICacheKey
    {
        string Create(HttpContext context, CommandCollection commands);
    }
}
```

And a default implementation that uses the relative path:

```c#
using Microsoft.AspNetCore.Http;
using Microsoft.AspNetCore.Http.Extensions;
using SixLabors.ImageSharp.Web.Commands;

namespace SixLabors.ImageSharp.Web.Caching
{
    public class RelativePathLowercaseCacheKey : ICacheKey
    {
        public string Create(HttpContext context, CommandCollection commands)
            => UriHelper.BuildRelative(context.Request.PathBase, context.Request.Path, QueryString.Create(commands)).ToLowerInvariant();
    }
}
```

Note: I'm already using the new `CommandCollection` that's part of another PR. And the implementation name should also indicate it's using a lowercased value, as some servers/OSes are case sensitive.

JimBobSquarePants:
If you just use the path that can (will absolutely because people are bad at naming things) increase the possibility of collision.

ronaldbarendse:
> If you just use the path that can (will absolutely because people are bad at naming things) increase the possibility of collision.

`ICacheKey` should only be concerted about _what_ makes the request unique (only the relative path, include hostname, username, etc.). `ICacheHash` is still responsible for ensuring that value doesn't have a high possibility of collision.

> I’m curious as to why you say the default setup/middleware doesn’t support multi tenancy because I’m using it out of the box with a multi tenant website and a **custom provider**.

Yes, so you had to use a custom provider to support multi tenancy (the default setup, using just `AddImageSharp()` doesn't allow that OOTB, e.g. by enabling some settings). If your custom provider uses something other than the URI to resolve to an image, that should be reflected in the cache key (which currently isn't the case).

In that sense, maybe the responsibility of returning the correct cache key should be added to `IImageProvider`? This implementation doesn't have to be aware of the commands though, so it would only return a unique cache key for the source image (e.g. `GetAsync(HttpContext context, out string cacheKey)`). The `ImageSharpMiddleware` should then add the commands to this key to make it unique for the processed image... I think that would make _even more_ sense and remove the need to add/set an additional `ICacheKey` implementation.

JimBobSquarePants:
You're more on the right track with your new interface idea, I really like it.

Although I think you're misunderstanding what I mean by collisions.

`ICacheHash` can and will only hash the input value. Let's consider a scenario where a naïve multitenancy implementation doesn't ensure that relative urls are unique even when the image should be (the inverse of the scenario in #145). 

```
[protocol][domainA]/path/image.jpg
[protocol][domainB]/path/image.jpg
```
This will generate the same hash since the relative paths are not unique. 

There are implementations like this, I've seen them, they are shockingly numerous, and they are hideous.

`IImageProvider` should never be responsible for returning a url. That would be a major violation of concern crossing. 

In my working implementation media item request use guids not slugified filenames to prevent collision and queries are prefiltered using EF Core query filters so I return true/false based ion the result of a query. It's a quick and easy way to handle multitenancy done right. 

So back to the interface. I wouldn't change `ICacheHash` because that introduces a string/uri/string conversion I'd like to avoid and as pointed out above we can heavily optimize our string creation here using `string.Create` since we know the exact size of the string we require. 

I would then create two implementations of `ICacheKey`. 

- `AbsolutePathCacheKey` (default so we don't break current cache implementations)
- `RelativePathCacheKey `

If I could turn back time I'd definitely use a relative path as the default (_naïve implementations be damned!_) but unfortunately I think we're stuck. There's breaking changes and then there's breaking changes. (_However, I could maybe be convinced as I don't believe anyone is paying for me to maintain this library_)

ronaldbarendse:
Creating different cache keys and thereby invalidating _all_ existing cached files would indeed be a functional change, especially because the cached files aren't automatically cleaned up! It wouldn't really break sites that upgrade though...

I've abstracted/implemented the existing cache key generation in PR https://github.com/SixLabors/ImageSharp.Web/pull/206 and also added some example/reference implementations.

If this gets released as part of v2, changing the default cache key implementation to use a lowercased relative URI wouldn't be a big issue, right? Upgrading will probably already require some code changes and adding a release note to advise users to manually clean up their image cache (or switch back to the v1 implementation) should be enough. Otherwise it will just re-generate the images and use up a bit more disk space (and potentially result in colliding with previously generated cache hashes for different images).

Regarding that last line: users should already always manually clean up their image cache after changing either the cache hash implementation anyway! So the same applies for changing the cache key implementation.

> If I could turn back time I'd definitely use a relative path as the default (_naïve implementations be damned!_) but unfortunately I think we're stuck. There's breaking changes and then there's breaking changes. (_However, I could maybe be convinced as I don't believe anyone is paying for me to maintain this library_)

Yeah, would be nice if you could go back and fix things before they bit you in the *ss 😆 Umbraco is sponsoring SixLabors on a monthly basis, although I guess that only pays for the coffee for the whole team 😅

JimBobSquarePants:
Newly merged #206 replaces this. Thanks everyone!

PR #148, opened 2021-03-23
Adapting log level when files are not found

### LINKED ISSUE #143: Wrong log level for non-existent files
Related to this log:

https://github.com/SixLabors/ImageSharp.Web/blob/451da152293154bff58c6943b37ca5b26ca9f26b/src/ImageSharp.Web/Middleware/LoggerExtensions.cs#L31

It's using Error level which means whenever a non-existent image is requested there will be a log message. We think (OC) this should be `Information` only.

I don't think we can lower the log level for specific events, and we don't want to discard all errors.

sebastienros:
Fixes #143

### Prerequisites

### Description
Error is too intrusive and these events are too common, the event log can then fill up quickly. Moving it to `Debug` since this is what aspnet uses for the standard static file provider on these events.


JimBobSquarePants:
Perfect. Thanks!

PR #154, opened 2021-04-08
Obsolete SetMemoryAllocator builder extensions

deanmarcussen:
### Prerequisites

### Description

Obsoletes `SetMemoryAllocator` (from https://github.com/SixLabors/ImageSharp.Web/discussions/152) - that's caused me confusion in the past before too :)

I was going to remove the comments, but then stylecop doesn't like it. 

Could update them to mention the obsolete message if preferable?

Could also just remove them, instead of obsolete...

### DIFF
--- a/src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
+++ b/src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
@@ -65,7 +65,7 @@ namespace SixLabors.ImageSharp.Web.DependencyInjection
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
         [Obsolete("Use ImageSharp.Configuration.MemoryAllocator. This will be removed in a future version.")]
-        public static IImageSharpBuilder SetMemoryAllocator<TMemoryAllocator>(this IImageSharpBuilder builder) 
+        public static IImageSharpBuilder SetMemoryAllocator<TMemoryAllocator>(this IImageSharpBuilder builder)
             where TMemoryAllocator : MemoryAllocator
             => builder;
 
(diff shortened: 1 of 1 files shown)


src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -65,7 +65,7 @@ public static IImageSharpBuilder SetMemoryAllocator(this IImageSharpBuilder buil
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
         [Obsolete("Use ImageSharp.Configuration.MemoryAllocator. This will be removed in a future version.")]
```
dlemstra, line 67: Should we make this an error instead? Otherwise I think we should keep the implementation to not break the usage?

deanmarcussen: as far as I can tell the current implementation would just register something in the DI, that is never retrieved, i.e. used.

So the only breaking, that I could think of with it, is if a user was using it as a convenience method to register a `TMemoryAllocator` and also then retrieving it from DI for their own purposes.

So here I just make it a noop instead.

But if you can think of another scenario?

dlemstra: I did not check the implementation details. Just noticed that the code was removed but would not break the build. I am still wondering if we should add `Error = true` to make sure the user changes their code.

deanmarcussen: Ah I get the suggestion now. Up to your preference really. 

The code that is removed didn't do anything (hasn't done for a long time I suspect), so it is questionably an error to call it.

I tend more towards not breaking builds on obsoleting where possible, so didn't go with error.

But as you prefer?

dlemstra: Will leave this up to @JimBobSquarePants to decide.

JimBobSquarePants:
It’ll be a few days til I can look at this. Sick at the moment with oral shingles.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
JimBobSquarePants: I have less sense than you both so would have just deleted it. :)

I doubt anyone is using this since it doesn't do anything other than provide a convenient way to register an allocator for their own usage. Lets make it an error.

deanmarcussen: Now I just want to delete it too ;)

Changed to `Error`

JimBobSquarePants (approved):

Issue #158, opened 2021-04-21
Response body is sometimes empty when processing multiple identical queries


markmccaigue:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp.Web
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
I'm brand new to the ImageSharp codebase, so apologies if this is user error.

Some requests seem to be getting responses with a 200 status code, but a ContentLength of 0 and an empty body. It seems this is mostly in cases where the same request is hitting the server a few times around the same sort of time.

### Steps to Reproduce
I've had a go at modifying one of the existing tests to assert the ContentLength is not 0, and it does fail for me, please see below.

AzureBlobStorageCacheServerTests
```
        [Theory]
        [InlineData(TestConstants.PhysicalTestImage)]
        [InlineData(TestConstants.AzureTestImage)]
        public async Task CanProcessMultipleIdenticalQueriesAsync(string url)
        {
            Task[] tasks = Enumerable.Range(0, 100).Select(i => Task.Run(async () =>
            {
                var command = i % 2 == 0 ? Command : Command2;
                using HttpResponseMessage response = await this.HttpClient.GetAsync(url + command);
                Assert.NotNull(response);
                Assert.True(response.IsSuccessStatusCode);
                // I have added this additional assertion below. If I modify the enumerable range to (0, 2), the test passes
                // If I leave the enumerable alone and generate a new width in each iteration, the test eventually passes
                // It seems to be only multiple identical requests which have the issue
                Assert.True(response.Content.Headers.ContentLength > 0);
            })).ToArray();

            var all = Task.WhenAll(tasks);
            await all;
            Assert.True(all.IsCompletedSuccessfully);
        }
```

I have a vague notion that we could be hitting imageContext.SendAsync fewer times than ProcessRequestAsync in the ImageSharpMiddleware, which makes me a little suspicious the issue could lie within some of the static caching in this middleware.

### System Configuration
The test above is failing for me in VS2019 16.9.3.

- ImageSharp.Web version:
[...]

PR #160, opened 2021-04-26
Fix Zero Length Images

deanmarcussen:
### Prerequisites

### Description

Fixes https://github.com/SixLabors/ImageSharp.Web/issues/158

As discussed on the issue the `ConcurrentDictionary` arrangement introduced to replace the `AsyncKeyLock` was misbehaving, and not always writing a response, when under high load. 

This pr rearranges the workers to always return an `ImageWorkResult` from which a response will be written.

It also replaces `Lazy` with what is recommended as a better way of controlling async `ConcurrentDictionary` `GetOrAdds`

Performance numbers drop unfortunately, but still remain better than the `AsyncKeyLock`
This pr : `Reqs/sec      6943.78`
Master : `Reqs/sec     41455.71`
AsyncKeyLock : `Reqs/sec      2299.19 ` (master at https://github.com/SixLabors/ImageSharp.Web/commit/aa1b1928c675cf02e86c3a6284b9d6b22d1880e5)

Note: Not rushing this, there's a couple of areas worth looking at a bit further, I will add some comments to the pr


src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -334,11 +344,27 @@ public async Task Invoke(HttpContext context)
                             // Save the image to the cache and send the response to the caller.
                             await this.cache.SetAsync(key, outStream, cachedImageMetadata);
 
-                            // Remove the resolver from the cache so we always resolve next request
+                            // Remove any resolver from the cache so we always resolve next request
                             // for the same key.
                             CacheResolverLru.TryRemove(key);
 
-                            await this.SendResponseAsync(imageContext, key, cachedImageMetadata, outStream, null);
```
deanmarcussen, line 341: This was essentially the issue, where the response was only sent once.

Because this factory now produces a result that must be sent multiple times, I've disposed of the `outStream` at the end of this block, and am returning a response from the cache that it creates. Otherwise it needs a way to keep track of how many awaiters are using the same stream (and lock the stream, during each response), so it can then finally dispose it.

Maybe an area to improve, but this seemed like the simpler solution, and it only occurs during a write

JimBobSquarePants:
>Performance numbers drop unfortunately, but still remain better than the AsyncKeyLock
This pr : Reqs/sec 6943.78
Master : Reqs/sec 41455.71
AsyncKeyLock : Reqs/sec 2299.19 (master at aa1b192)

Ooft that stings! Better a working slower solution but ow!

deanmarcussen:
> Ooft that stings! Better a working slower solution but ow!

I am gutted :(

JimBobSquarePants:
I'm gonna have a deeper look at the issue here. I'm wondering if there's a less drastic fix.

deanmarcussen:
That would be great @JimBobSquarePants this needs as many eyes as possible.

I've had a final run through with this pr, and am happy with it.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
JimBobSquarePants: I'm still trying to understand the cause of our issue. We're only returning the metadata as the result of our Lazy so it should be safe to return the result multiple times since we're always requesting the stream from that metadata via a method. (we don't keep hold of the stream)

https://github.com/SixLabors/ImageSharp.Web/blob/8f67aae9d314a6fde96aae856027a655d6cbd2fa/src/ImageSharp.Web/Resolvers/PhysicalFileSystemResolver.cs#L33

Are we holding onto the result for too long?

deanmarcussen: The issue is that in the current Lazy read worker is is only returning whether to refresh the disccache or not. 
It's actually doing the send of the image response inside the worker valuefactory.
And because the point of the lazy was to stop those valuefactories running multiple times, it was doing exactly that. Running it once. So 5 awaiters would await the result, but all the result would do is return or allow a refresh of the cached image. 

```
            (bool newOrUpdated, ImageMetadata sourceImageMetadata) =
                await this.IsNewOrUpdatedAsync(sourceImageResolver, imageContext, key); <-- the sendresponse happens once. no matter how many awaiters

            if (!newOrUpdated)
            {
                return;
            }
```

so the change is to get a cacheresolver back from the task.
then all awaiters can send the image response, each awaiter, opening the `CreateReadStream()`

src/ImageSharp.Web/Middleware/ImageWorkerResult.cs
```diff
@@ -0,0 +1,45 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+using SixLabors.ImageSharp.Web.Resolvers;
+
+namespace SixLabors.ImageSharp.Web.Middleware
+{
+    /// <summary>
+    /// Provides an asynchronous worker result.
+    /// </summary>
+    internal struct ImageWorkerResult
```
JimBobSquarePants, line 11: This should be a readonly struct with any methods accepting the type as a parameter passed via `in` to prevent a copy.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
JimBobSquarePants: Ah ok.. So our numbers were junk so the loss avoidable. 

Just to clarify the corrected behavior.

Given A-C are unique URLs numbered 1-2

Request A (1) is received - This is **uncached** but is not blocked because there are no current workers processing that URL.
Request B (1) is received - This is **uncached** but is not blocked because there are no current workers processing that URL.
Request C (1) is received  - This is **cached** but is not blocked because there are no current workers processing that URL.
Request C (2) is received  - This is **cached** but is not blocked because C (1) already processing that URL is a cache read.
Request C (1) is complete.
Request C (2) is complete.
Request A (2) is received - This is **blocked** because A (1) is a write request and has not completed.
Request B (2) is received - This is **blocked** because B (1) is a write request and has not completed.
Request A (1) is complete.
Request A (2) is complete and reads from the newly cached file created by A (1).
Request B (1) is complete.
Request B (2) is complete and reads from the newly cached file created by B (1).

src/ImageSharp.Web/Middleware/ImageWorkerResult.cs
deanmarcussen: Done. It's only a holder, so not passed anywhere that can take an `in`

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
deanmarcussen: > Just to clarify the corrected behavior.

Yes, that adds up to me.

The only comment I would make is just to clarify that the read workers still block/await
- Request C (1) is received - This is cached but is not blocked because there are no current workers processing that URL.
- Request C (2) is received - This is cached but is not blocked because C (1) already processing that URL is a cache read.

In this scenario C (1) is cached on disc, but not yet in LRU. C (1) blocks the LRU read (read of source metadata, and read of cache metadata). When it completes C (2) is also freed to return the same LRU result. (and when C (3) arrives later, it gets the LRU result, without doing the metadata reads).

I looked to see if the read worker could be removed (just to simplify), and the LRU Cache made so it's `GetOrAddAsync` could be blocked with the same technique, but because there is a source metadata read, and cache metadata read to do, which are in different LRU caches, it wasn't viable, and better to block the initial reads into the both of the LRU caches.

JimBobSquarePants: Yep that is sensible and works for me. I think we’re good to go with this, nice job!

JimBobSquarePants (approved):

mariojsnunes:
Hi, when can we expect a release with this? The memory leak fix is really important :)

JimBobSquarePants:
I’m awaiting confirmation in the discussion here from a support license holder.

https://github.com/SixLabors/ImageSharp.Web/discussions/161

deanmarcussen:
@mariojsnunes the release is available on the myget feed https://www.myget.org/feed/sixlabors/package/nuget/SixLabors.ImageSharp.Web/1.0.3-alpha.0.4

It would be great to get any feedback on it.

PR #163, opened 2021-04-29
Dispose of the cached stream.

### LINKED ISSUE #162: Cache image reponses are not disposed
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Web
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description

There is a couple of reported memory related issues, and while I know there is some tracking in ImageSharp itself about releasing retained resources, there looks to be another issue here (easily resolved) regarding stream usage.

Reported here on Orchard Core https://github.com/OrchardCMS/OrchardCore/issues/9060
and likely related to this discussion https://github.com/SixLabors/ImageSharp.Web/discussions/161

Performance testing locally this causes memory usage to grow dramatically - got as high as 15Gb before I stopped it.

Collecting manually doesn't drop it, so I suspect that's why there's reported app service failures.

### Steps to Reproduce

This line doesn't dispose of the cache file streams
https://github.com/SixLabors/ImageSharp.Web/blob/8f67aae9d314a6fde96aae856027a655d6cbd2fa/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs#L476

adding a dispose
```
                    using (var stream = await cacheResolver.OpenReadAsync())
                    {
                        await imageContext.SendAsync(stream, metadata);
                    }
```

and everything gets collected when the GC runs, and memory usage is as one would expect.

I've seen some (difficult to reproduce) issues with file locks on the cache images, but I am hoping this might explain those as well.

Off for the covid jab in a minute, so pr to come

### System Configuration

Just locally.

JimBobSquarePants:
### Prerequisites

### Description

Ensures that the cached image stream is disposed of.   Fixes #162

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -473,7 +473,11 @@ namespace SixLabors.ImageSharp.Web.Middleware
                     this.logger.LogImageServed(imageContext.GetDisplayUrl(), key);
 
                     // When stream is null we're sending from the cache.
-                    await imageContext.SendAsync(stream ?? await cacheResolver.OpenReadAsync(), metadata);
+                    using (Stream output = stream ?? await cacheResolver.OpenReadAsync())
+                    {
+                        await imageContext.SendAsync(output, metadata);
+                    }
+
                     return;
 
                 case ImageContext.PreconditionState.NotModified:


src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -473,7 +473,11 @@ private ValueTask StreamDisposeAsync(Stream stream)
                     this.logger.LogImageServed(imageContext.GetDisplayUrl(), key);
 
                     // When stream is null we're sending from the cache.
-                    await imageContext.SendAsync(stream ?? await cacheResolver.OpenReadAsync(), metadata);
+                    using (Stream output = stream ?? await cacheResolver.OpenReadAsync())
```
dlemstra, line 476: This will also mean that you will dispose the input stream. Don't think we should?

JimBobSquarePants: Ah no, we always copy the input stream so we have control.

https://github.com/SixLabors/ImageSharp.Web/blob/8f67aae9d314a6fde96aae856027a655d6cbd2fa/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs#L296-L302

dlemstra: But later we will call `StreamDisposeAsync`?

JimBobSquarePants: Yeah but I'd consider that safe. `Dispose`, by design should allow multiple calls and `RecyclableMemoryStream` certainly allows it. The pipeline is going through a bit of a refactor just now via #160 so it's gonna change anyway.

deanmarcussen: If the pipeline was working how it should this would be an issue, but it’s not, so that outsream is single use, so as long as it handles dispose being called twice ok it should be ok

deanmarcussen: Lot of shoulds in that statement :)

 makes me think it might be better to handle the two types of streams separately, at least until the refactor removes it

JimBobSquarePants:
Closed in favour of #160

Issue #182, opened 2021-08-11
Processing does not use query string parameter order


ronaldbarendse:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have verified that I am running the latest version of ImageSharp.Web
- [x] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [x] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### Description
According to the [documentation](https://docs.sixlabors.com/articles/imagesharp.web/processingcommands.html) processing operations should use the order in which they are supplied in the query string, but this is currently not the case. Instead, it loops through all processors in the order they are registered and calls the `Process` method on each one (even when none of the commands for that processor are provided):

https://github.com/SixLabors/ImageSharp.Web/blob/d0559520245dffc484cd7fc77ae35fd20f0ff3b9/src/ImageSharp.Web/Processors/WebProcessingExtensions.cs#L28-L45

This probably wasn't noticed before, because the order in which the built-in processors are called doesn't affect the end-result (`ResizeWebProcessor`, `FormatWebProcessor`, `BackgroundColorWebProcessor` and `JpegQualityWebProcessor` all affect different aspects of an image). All processors also include a check before mutating the image, so that ensured there were no side-effects by always invoking them.

Umbraco required a custom processor to crop images using coordinates (representing a crop box/rectangle), but adding that after the default built-in processors would result in cropping the already resized image, instead of using the original.

### Steps to Reproduce
You can use the `CropWebProcessor` from https://github.com/umbraco/Umbraco-CMS/pull/10623 to test this issue (or create a similar simple one):
- Only adding the processor (after the default ones) ends up with wrongly cropped/resized images;
- Removing the `ResizeWebProcessor`, adding the custom processor and then re-adding it again will correctly crop before resizing.

### Suggested solution
The `ImageSharpMiddleware` already keeps track of all known processing commands, filters unknown ones from the current request and short-circuits when there are none. This would mean only the above `Process` extension method will need to be updated.

Besides invoking the processor in the order they are provided in the query string, we need to ensure they aren't invoked for every command (e.g. twice when using `?width=200&height=200`). This can be done by tracking the all commands used by a processor:

```c#
var processedCommands = new List<string>();
[...]

PR #183, opened 2021-08-30
Order and filter processors by supported commands

ronaldbarendse:
### Prerequisites

### Description
This fixes issue https://github.com/SixLabors/ImageSharp.Web/issues/182 by ordering and filtering the processors by supported commands.

The biggest change this introduces is not calling the `Process(...)` method on processors where the `Commands` collection doesn't contain a parsed command.

### DIFF
--- a/src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
+++ b/src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
 using Microsoft.Extensions.Logging;
 using SixLabors.ImageSharp.Web.Commands;
 
@@ -33,15 +34,27 @@ namespace SixLabors.ImageSharp.Web.Processors
             CommandParser commandParser,
             CultureInfo culture)
         {
-            if (commands.Count != 0)
+            foreach (IImageWebProcessor processor in processors.GetBySupportedCommands(commands.Keys.ToList()))
             {
-                foreach (IImageWebProcessor processor in processors)
-                {
-                    source = processor.Process(source, logger, commands, commandParser, culture);
-                }
+                source = processor.Process(source, logger, commands, commandParser, culture);
             }
 
             return source;
         }
+
+        /// <summary>
+        /// Sorts the processors according to the first supported command and removes processors without supported commands.
+        /// </summary>
+        /// <param name="processors">The collection of available processors.</param>
+        /// <param name="commands">The list of parsed commands.</param>
+        /// <returns>
+        /// The sorted proccessors.
+        /// </returns>
+        public static IEnumerable<IImageWebProcessor> GetBySupportedCommands(this IEnumerable<IImageWebProcessor> processors, IList<string> commands)
+            => processors
+                .GroupBy(p => p.Commands.Min(c => (int?)commands.IndexOf(c)) ?? -1) // Get index of first supported command
+                .Where(g => g.Key != -1) // Remove processors without supported commands
+                .OrderBy(g => g.Key)
+                .SelectMany(g => g);
     }
 }


src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
```diff
@@ -33,15 +34,27 @@ internal static class WebProcessingExtensions
...
+        /// Sorts the processors according to the first supported command and removes processors without supported commands.
+        /// </summary>
+        /// <param name="processors">The collection of available processors.</param>
+        /// <param name="commands">The list of parsed commands.</param>
+        /// <returns>
+        /// The sorted proccessors.
+        /// </returns>
+        public static IEnumerable<IImageWebProcessor> GetBySupportedCommands(this IEnumerable<IImageWebProcessor> processors, IList<string> commands)
+            => processors
+                .GroupBy(p => p.Commands.Min(c => (int?)commands.IndexOf(c)) ?? -1) // Get index of first supported command
```
ronaldbarendse, line 55: We don't expect/support `Commands` to return `null`, but it could return an empty collection: in that case we fallback to -1 and filter out the processor. This means it's not possible to create a processor that is called for all requests.

We could default to 0 (that would run it as one of the first processors though) or introduce a new property on the `IImageWebProcessor` to provide this value, any thoughts?

src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
```diff
@@ -33,15 +34,27 @@ internal static class WebProcessingExtensions
...
+                .Where(g => g.Key != -1) // Remove processors without supported commands
+                .OrderBy(g => g.Key)
```
ronaldbarendse, line 57:
```suggestion
                .Where(g => g.Key != -1) // Filter processors without supported commands
                .OrderBy(g => g.Key) // Order processors by first supported command
```

ronaldbarendse:
I just tested this PR again (with browser caching off) and noticed it always returned -1 as lowest index... Now it uses an intersection of parsed commands and commands supported by the processor, so it really gets the first index _of a supported command_ 😄

Because case sensitivity of the commands depend on the comparer used in the dictionary returned by `ParseRequestCommands`, I've made sure this is also used when stripping out unknown commands and getting the first index for processing. Previously when requesting `/imagesharp-logo.png?WIDTH=300` (in Edge, as Chrome seems to be smart and remember the 'correct' URL casing), it stripped out the command, even though the `ResizeWebProcessor` would use the correct width/value.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -188,10 +188,12 @@ public async Task Invoke(HttpContext context)
...
             if (commands.Count > 0)
             {
-                // Strip out any unknown commands.
+                // Strip out any unknown commands
+                IEqualityComparer<string> comparer = (commands as Dictionary<string, string>)?.Comparer ?? StringComparer.OrdinalIgnoreCase;
+
                 foreach (string command in new List<string>(commands.Keys))
                 {
-                    if (!this.knownCommands.Contains(command))
+                    if (!this.knownCommands.Contains(command, comparer))
```
ronaldbarendse, line 196: I'm also fine with removing the above line and just using:
```suggestion
                    if (!this.knownCommands.Contains(command, StringComparer.OrdinalIgnoreCase))
```

src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
```diff
@@ -33,15 +35,32 @@ internal static class WebProcessingExtensions
...
+        /// <returns>
+        /// The sorted proccessors.
+        /// </returns>
+        public static IEnumerable<IImageWebProcessor> GetBySupportedCommands(this IEnumerable<IImageWebProcessor> processors, IDictionary<string, string> commands)
+        {
+            List<string> commandKeys = new List<string>(commands.Keys);
+            IEqualityComparer<string> comparer = (commands as Dictionary<string, string>)?.Comparer ?? StringComparer.OrdinalIgnoreCase;
+
+            return processors
+                .GroupBy(p => commandKeys.Intersect(p.Commands, comparer).Min(c => (int?)commandKeys.FindIndex(k => comparer.Equals(c, k))) ?? -1) // Get index of first supported command
```
ronaldbarendse, line 60: Similar to `ImageSharpMiddleware`, we can also remove the above `comparer` and just use:
```suggestion
                .GroupBy(p => commandKeys.Intersect(p.Commands, StringComparer.OrdinalIgnoreCase).Min(c => (int?)commandKeys.FindIndex(k => k.Equals(c, StringComparison.OrdinalIgnoreCase))) ?? -1) // Get index of first supported command
```

If we do this, we might want to change the `commands` parameter type into a list, as we don't need the dictionary/values in this method anyway...

src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
ronaldbarendse: Lets not complicate things any further (and introduce a breaking change by adding a new property to the `IImageWebProcessor`): if you want a processor to be called on every request, you can add the command in the `OnParseCommandsAsync` callback, in a custom `IRequestParser` or even manually do any processing in the `OnProcessedAsync` callback.

src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
```diff
@@ -33,15 +35,29 @@ internal static class WebProcessingExtensions
...
+        /// Sorts the processors according to the first supported command and removes processors without supported commands.
+        /// </summary>
+        /// <param name="processors">The collection of available processors.</param>
+        /// <param name="commands">The parsed collection of processing commands.</param>
+        /// <returns>
+        /// The sorted proccessors that supports any of the specified commands.
+        /// </returns>
+        public static IEnumerable<IImageWebProcessor> GetBySupportedCommands(this IEnumerable<IImageWebProcessor> processors, List<string> commands)
+            => processors
+                .GroupBy(p => commands.FindIndex(c => p.Commands.Contains(c, StringComparer.OrdinalIgnoreCase))) // Get index of first supported command
```
ronaldbarendse, line 58: I've greatly simplified this check (compared to the first commits) and although `FindIndex()` is an O(n) operation, unknown commands are already stripped and this will only be run when the image needs to get processed (i.e. is not already cached).

JimBobSquarePants: Can't have Linq on a hot path like this.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -188,10 +188,10 @@ public async Task Invoke(HttpContext context)
             IDictionary<string, string> commands = this.requestParser.ParseRequestCommands(context);
             if (commands.Count > 0)
             {
-                // Strip out any unknown commands.
+                // Strip out any unknown commands
                 foreach (string command in new List<string>(commands.Keys))
                 {
-                    if (!this.knownCommands.Contains(command))
+                    if (!this.knownCommands.Contains(command, StringComparer.OrdinalIgnoreCase))
```
JimBobSquarePants, line 194: I actually want command case to be respected. Reduces attack surface

JimBobSquarePants (commented):
Thanks for the PR. I need to do a thorough review + perf check this but I've made some notes.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
ronaldbarendse: The ASP.NET Core runtime uses case insensitive query string parsing though (check the result of `Microsoft.AspNetCore.WebUtilities.QueryHelpers.ParseQuery("?test=true").Comparer` or https://github.com/dotnet/aspnetcore/issues/3450) and all the `IRequestParser` implementations in ImageSharp also return a case insensitive dictionary...

Besides that, the cache hash is already generated based on a [lower cased URI](https://github.com/SixLabors/ImageSharp.Web/blob/fa183578c358c63f3225dd76b3d0a92413f038f8/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs#L538), so I don't see how this would reduce the attack surface.

src/ImageSharp.Web/Processors/WebProcessingExtensions.cs
ronaldbarendse: I'll rewrite this into a more verbose, non-LINQ, version 👍🏻

PR #185, opened 2021-10-04
Use all matching/valid image providers to get the image resolver

ronaldbarendse:
### Prerequisites

### Description
The `PhysicalFileSystemProvider` is currently added by default (and thus as first) and matches all requests, which results in the `ImageSharpMiddleware` only considering physical files (or at least only the `WebRootFileProvider`) and not fallback to other providers.

This is an issue, because you can't change the order of registered providers (they are returned in the order which they are registered into DI) and it makes sense to first check the local filesystem before using cloud storage providers like the [`AzureBlobStorageImageProvider`](https://docs.sixlabors.com/articles/imagesharp.web/imageproviders.html#azureblobstorageimageprovider). The documentation and API also doesn't help much, as it only says _"Url matching for this provider follows the same rules as conventional static files."_ and `services.AddImageSharp()` already adds the default `PhysicalFileSystemProvider`, which you would then first need to remove and add back again after registering e.g. the `AzureBlobStorageImageProvider`.

On top of this, the web root can also serve static files using a `CompositeFileProvider`, which would mean the 'normal' static files can use a fallback, but ImageSharp doesn't support this (at least in an easy way).

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -9,7 +9,6 @@ using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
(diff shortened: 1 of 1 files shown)


JimBobSquarePants (commented):
I’m afraid I’m going to have to close this. As I described elsewhere the current process is as design. You really should raise discussions before PRs.

PR #188, opened 2021-11-12
Avoid most allocations in GetExtensionFromUri

PhyxionNL:
### Prerequisites

### Description
Avoid some of the string allocations in GetExtensionFromUri, and it's also faster.

| Method |     Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|------- |---------:|---------:|---------:|------:|--------:|-------:|----------:|
|    Old | 74.75 us | 0.813 us | 0.761 us |  1.00 |    0.00 | 2.0752 |     18 KB |
|    New | 60.40 us | 1.139 us | 1.119 us |  0.81 |    0.02 | 1.2207 |     11 KB |
|    NewV2 | 57.30 us | 0.905 us | 0.846 us |  0.77 | 0.02 | 0.9766 |      8 KB |
|    NewV3 | 56.38 us | 0.348 us | 0.325 us |  0.77 | 0.02 |  0.3662 |      3 KB |

### DIFF
--- a/src/ImageSharp.Web/FormatUtilities.cs
+++ b/src/ImageSharp.Web/FormatUtilities.cs
@@ -65,19 +65,11 @@ namespace SixLabors.ImageSharp.Web
                     return ext;
                 }
 
-#if !NETCOREAPP3_1_OR_GREATER
-                path = uri.ToLowerInvariant().AsSpan(0, query);
-#else
                 path = uri.AsSpan(0, query);
-#endif
             }
             else
             {
-#if !NETCOREAPP3_1_OR_GREATER
-                path = uri.ToLowerInvariant();
-#else
                 path = uri;
-#endif
             }
 
             int extensionIndex;
--- a/src/ImageSharp.Web/FormatUtilities.cs
+++ b/src/ImageSharp.Web/FormatUtilities.cs
@@ -19,31 +19,27 @@ namespace SixLabors.ImageSharp.Web
     /// </summary>
     public sealed class FormatUtilities
     {
-        private readonly List<string> fileExtensions = new List<string>();
-        private readonly Dictionary<string, string> fileExtension = new Dictionary<string, string>();
+        private readonly List<string> extensions = new();
+        private readonly Dictionary<string, string> extensionsByMimeType = new();
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="FormatUtilities"/> class.
+        /// Initializes a new instance of the <see cref="FormatUtilities" /> class.
         /// </summary>
         /// <param name="options">The middleware options.</param>
         public FormatUtilities(IOptions<ImageSharpMiddlewareOptions> options)
         {
             Guard.NotNull(options, nameof(options));
 
-            // The formats contained in the configuration are used a lot in hash generation
-            // so we need them to be enumerated to remove allocations and allow indexing.
-            IImageFormat[] imageFormats = options.Value.Configuration.ImageFormats.ToArray();
-
-            for (int i = 0; i < imageFormats.Length; i++)
+            foreach (IImageFormat imageFormat in options.Value.Configuration.ImageFormats)
             {
-                string[] extensions = imageFormats[i].FileExtensions.ToArray();
+                string[] extensions = imageFormat.FileExtensions.ToArray();
 
                 foreach (string extension in extensions)
                 {
-                    this.fileExtensions.Add(extension);
+                    this.extensions.Add(extension);
                 }
 
-                this.fileExtension[imageFormats[i].DefaultMimeType] = extensions[0];
+                this.extensionsByMimeType[imageFormat.DefaultMimeType] = extensions[0];
             }
         }
 
@@ -51,7 +47,7 @@ namespace SixLabors.ImageSharp.Web
         /// Gets the file extension for the given image uri.
         /// </summary>
         /// <param name="uri">The full request uri.</param>
-        /// <returns>The <see cref="string"/>.</returns>
+        /// <returns>The <see cref="string" />.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public string GetExtensionFromUri(string uri)
         {
@@ -75,13 +71,13 @@ namespace SixLabors.ImageSharp.Web
             int extensionIndex;
             if ((extensionIndex = path.LastIndexOf('.')) != -1)
             {
-                ReadOnlySpan<char> extension = path.Slice(extensionIndex + 1);
+                ReadOnlySpan<char> pathExtension = path.Slice(extensionIndex + 1);
 
-                foreach (string ext in this.fileExtensions)
+                foreach (string extension in this.extensions)
                 {
-                    if (extension.Equals(ext, StringComparison.OrdinalIgnoreCase))
+                    if (pathExtension.Equals(extension, StringComparison.OrdinalIgnoreCase))
                     {
-                        return ext;
+                        return extension;
                     }
                 }
             }
@@ -93,8 +89,8 @@ namespace SixLabors.ImageSharp.Web
         /// Gets the correct extension for the given content type (mime-type).
         /// </summary>
         /// <param name="contentType">The content type (mime-type).</param>
-        /// <returns>The <see cref="string"/>.</returns>
+        /// <returns>The <see cref="string" />.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public string GetExtensionFromContentType(string contentType) => this.fileExtension[contentType];
+        public string GetExtensionFromContentType(string contentType) => this.extensionsByMimeType[contentType];
     }
 }
--- a/src/ImageSharp.Web/FormatUtilities.cs
+++ b/src/ImageSharp.Web/FormatUtilities.cs
@@ -19,8 +19,7 @@ namespace SixLabors.ImageSharp.Web
     /// </summary>
     public sealed class FormatUtilities
     {
-        private readonly IImageFormat[] imageFormats;
-        private readonly Dictionary<IImageFormat, string[]> fileExtensions = new Dictionary<IImageFormat, string[]>();
+        private readonly List<string> fileExtensions = new List<string>();
         private readonly Dictionary<string, string> fileExtension = new Dictionary<string, string>();
 
         /// <summary>
@@ -33,12 +32,18 @@ namespace SixLabors.ImageSharp.Web
 
             // The formats contained in the configuration are used a lot in hash generation
             // so we need them to be enumerated to remove allocations and allow indexing.
-            this.imageFormats = options.Value.Configuration.ImageFormats.ToArray();
-            for (int i = 0; i < this.imageFormats.Length; i++)
+            IImageFormat[] imageFormats = options.Value.Configuration.ImageFormats.ToArray();
+
+            for (int i = 0; i < imageFormats.Length; i++)
             {
-                string[] extensions = this.imageFormats[i].FileExtensions.ToArray();
-                this.fileExtensions[this.imageFormats[i]] = extensions;
-                this.fileExtension[this.imageFormats[i].DefaultMimeType] = extensions[0];
+                string[] extensions = imageFormats[i].FileExtensions.ToArray();
+
+                foreach (string extension in extensions)
+                {
+                    this.fileExtensions.Add(extension);
+                }
+
+                this.fileExtension[imageFormats[i].DefaultMimeType] = extensions[0];
             }
         }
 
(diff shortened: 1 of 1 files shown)


PhyxionNL:
Added a second commit to remove most allocations from this method if running .NET Core 3.1 and higher. Maybe drop .NET Core 2.1 target? It's end of life.

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -20,6 +20,7 @@ namespace SixLabors.ImageSharp.Web
     public sealed class FormatUtilities
     {
         private readonly IImageFormat[] imageFormats;
+        private readonly Dictionary<IImageFormat, string[]> dottedFileExtensions = new Dictionary<IImageFormat, string[]>();
```
sebastienros, line 23: Maybe worth indexing on the `IImageFormat.Name.ToLowerInvariant()` as it will make a better index for perf. No dynamic resolution of `GetHashCode()` and `Equals()`. Valid for all indexes here.

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -50,21 +52,39 @@ public FormatUtilities(IOptions<ImageSharpMiddlewareOptions> options)
...
+#if !NETCOREAPP3_1_OR_GREATER
+            uri = uri.ToLowerInvariant();
+#endif
+
+            int query = uri.IndexOf('?');
+            ReadOnlySpan<char> path;
+
+            if (query > -1)
+            {
+                if (QueryHelpers.ParseQuery(uri.Substring(query)).TryGetValue(FormatWebProcessor.Format, out StringValues ext))
```
sebastienros, line 64: Parsing can be skipped if the string doesn't contain `FormatWebProcessor.Format.ToLowerInvariant()`, might be worth it since otherwise it's creating a Dictionary on every call, and a new string with `Substring()`.

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -50,21 +52,39 @@ public FormatUtilities(IOptions<ImageSharpMiddlewareOptions> options)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public string GetExtensionFromUri(string uri)
         {
-            // TODO: Investigate using span to reduce allocations here.
-            string[] parts = uri.Split('?');
-            if (parts.Length > 1 && QueryHelpers.ParseQuery(parts[1]).TryGetValue(FormatWebProcessor.Format, out StringValues ext))
+#if !NETCOREAPP3_1_OR_GREATER
+            uri = uri.ToLowerInvariant();
```
sebastienros, line 56: Not necessary if it goes in the `ParseQuery` section

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -50,21 +52,39 @@ public FormatUtilities(IOptions<ImageSharpMiddlewareOptions> options)
...
+                path = uri;
             }
 
-            string path = parts[0];
             string extension = null;
             int index = 0;
             for (int i = 0; i < this.imageFormats.Length; i++)
             {
-                for (int j = 0; j < this.fileExtensions[this.imageFormats[i]].Length; j++)
+                string[] extensions = this.dottedFileExtensions[this.imageFormats[i]];
```
sebastienros, line 80: `this.imageFormats[i]` can be memoized as it's reused later

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -50,21 +52,39 @@ public FormatUtilities(IOptions<ImageSharpMiddlewareOptions> options)
...
+                for (int j = 0; j < extensions.Length; j++)
                 {
-                    int li = path.LastIndexOf($".{this.fileExtensions[this.imageFormats[i]][j]}", StringComparison.OrdinalIgnoreCase);
+#if NETCOREAPP3_1_OR_GREATER
+                    int li = path.LastIndexOf(extensions[j], StringComparison.OrdinalIgnoreCase);
+#else
+                    int li = path.LastIndexOf(extensions[j]);
```
sebastienros, line 86: Instead of checking for every possible match, why isn't it extracting the actual extension, and then doing a single lookup over a dictionary?

src/ImageSharp.Web/FormatUtilities.cs
PhyxionNL: Thanks 👍 I had just picked some low hanging fruit. I have changed this 🙂

src/ImageSharp.Web/FormatUtilities.cs
PhyxionNL: I think this is done to make it more resistant to files with additional extensions (.png.tmp for example), although I'm not sure how often that happens, and it's not ideal for performance. If you just extract the extension and use a lookup for it:

|    NewV4 | 9.814 us | 0.0426 us | 0.0398 us | 0.4730 |      4 KB |

But I think we need to know first ( @JimBobSquarePants ) why it was done like this 🙂

PhyxionNL: Although if you want to retain that behavior you can loop through extensions from the end, until you find one that's supported.  Takes a small hit though.

| Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------- |---------:|---------:|---------:|-------:|----------:|
|    NewV5 | 12.13 us | 0.091 us | 0.085 us | 0.7477 |      6 KB |

JimBobSquarePants: We check for the '.' prefix as that excludes any filenames containing strings that match extension names. e.g. `mybmp.jpg`

PhyxionNL: OK, then it's not needed to check for multiple extensions (which seems rather pointless anyway), so the NewV4 variant can be used (with 1KB less allocation on .NET Core 3.1+), which would further cut down the time ~6x. Let me commit that version.

PhyxionNL: Committed, tests still pass 🙂

JimBobSquarePants: If you introduce the new `dottedFileExtensions` dictionary (_no need to use anything other than the default comparer_) as you have and use a string for the key for `fileExtensions` the entire method can become as simple as this.

```c#
public string GetExtensionFromUri(string uri)
{
    int queryIndex = uri.IndexOf('?');
    if (queryIndex != -1 && QueryHelpers.ParseQuery(uri.Substring(queryIndex)).TryGetValue(FormatWebProcessor.Format, out StringValues ext))
    {
        return ext;
    }

    string extension = null;
    int index = 0;
    for (int i = 0; i < this.imageFormats.Length; i++)
    {
        string name = this.imageFormats[i].Name;
        string[] extensions = this.fileExtensions[name];
        string[] dottedExtensions = this.dottedFileExtensions[name];
        for (int j = 0; j < dottedExtensions.Length; j++)
        {
            int li = uri.LastIndexOf(dottedExtensions[j], StringComparison.OrdinalIgnoreCase);
            if (li < index)
            {
                continue;
            }

            index = li;
            extension = extensions[j];
        }
    }

    return extension;
}
```

JimBobSquarePants: >OK, then it's not needed to check for multiple extensions (which seems rather pointless anyway)

No you still need to check for both scenarios. Someone could appear with `mybmp.jpg` and `mybmp.png.jpg`

PhyxionNL: > 
> 
> > OK, then it's not needed to check for multiple extensions (which seems rather pointless anyway)
> 
> No you still need to check for both scenarios. Someone could appear with `mybmp.jpg` and `mybmp.png.jpg`

What would you expect the format to be in this case? png or jpg? I would expect jpg, which would be valid with the code as-is.

The code above is much slower than simply extracting the extension.

PhyxionNL: BTW, if you want to loop through everything, it's trivial to change the if to a while and add ```path = path.Slice(0, extensionIndex);``` at the end of the while. Then it will loop from end to start.

JimBobSquarePants: I'm not sure using the hashcode as you have is safe here. 
https://stackoverflow.com/a/7969189/427899

PhyxionNL: It's just over the extension and I've checked, there are no collisions there. The post is from 2011 and in .NET Core the hashcode algorithm changed to Marvan and no longer collides so much :)

JimBobSquarePants: What's the cost of `Substring` nowadays? Has that been optimized?

PhyxionNL: Considerably, it's the difference between NewV6 and NewV2.

JimBobSquarePants: Ah no, I don't mean a collision between extensions I meant a collision between the hashed sliced input value and a proper extension. The probability of getting a string with a dot and following value that collided would be crazy small though! :)

I'm wondering if we can avoid the dictionary? I have an Idea I'd like to run past you?

Let's say that instead of using a dictionary `fileExtensions` was instead an array containing of all the combined extensions from all image formats. Stealing a lot of the clever stuff you've done already we could slice the input uri and do a lookup against that array. `ToString()` on a `ReadonlySpan<char>` costs nothing and the ordinal comparison should be fast enough. What do you think?

```c#
public string GetExtensionFromUri(string uri)
{
    ReadOnlySpan<char> path;
    int queryIndex = uri.IndexOf('?');
    if (queryIndex != -1)
    {
        if (QueryHelpers.ParseQuery(uri.Substring(queryIndex)).TryGetValue(FormatWebProcessor.Format, out StringValues ext))
        {
            return ext;
        }

        path = uri.AsSpan(0, queryIndex);
    }
    else
    {
        path = uri;
    }

    int index;
    if ((index = path.LastIndexOf('.')) != -1)
    {
        uri = path.Slice(index + 1).ToString();
        string[] extensions = this.fileExtensions;
        for (int i = 0; i < extensions.Length; i++)
        {
            string extension = extensions[i];
            if (extension.Equals(uri, StringComparison.OrdinalIgnoreCase))
            {
                return extension;
            }
        }
    }

    return null;
}
```

sebastienros: Please don't index on a hashcode, it doesn't make sense (to me), both for correctness and performance.

When you think an iteration is faster than a lookup, use an `HybridDictionary`, it does it for you, converts to an actual dictionary if there are many elements.

Also the `.ToString()` "will" allocate, obviously. You can do the same comparison allocation-free by only using slices and span comparisons on the uri (endswith) if you really don't want to use a dictionary. Then I would suggest to order the extensions by MRU since it's likely that the same extensions are used over and over. And if not MRU as it might get costly, have two collections, "used" and "never-used", then it might be fine iterating on the "used" collections, and doing a lookup on the "never-used" as a fallback.

All this is very fun, more fun that the actual gains ;)

PhyxionNL: Thanks @sebastienros. I have a page with a lot of images on it and this takes the majority of the page load time. Not so noticeable on my dev machine, but on a slow machine it takes quite some time as this gets called per image request.
@JimBobSquarePants 
I have changed the implementation to use a List and loop through it, this would still be ~6x as fast as the original implementation.

| Method |     Mean |    Error |   StdDev |  Gen 0 | Allocated |
|------- |---------:|---------:|---------:|-------:|----------:|
|    New | 11.36 us | 0.206 us | 0.182 us | 0.3967 |      3 KB |

sebastienros: > I have a page with a lot of images on it

What I meant is the number of distinct extensions, even if you have thousands of images on this page, they probably are using the same extension. I would also assume most extensions used in an app are always the same, like `.png`, `.jpg`, ... so splitting the list of extensions based on usage might be a better optimization since the main list would then have just a few extensions in it, and a second list have the remaining ones.

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -50,32 +55,46 @@ public FormatUtilities(IOptions<ImageSharpMiddlewareOptions> options)
...
+#if !NETCOREAPP3_1_OR_GREATER
+                path = uri.ToLowerInvariant().AsSpan(0, query);
+#else
+                path = uri.AsSpan(0, query);
+#endif
+            }
+            else
+            {
+#if !NETCOREAPP3_1_OR_GREATER
+                path = uri.ToLowerInvariant();
```
sebastienros, line 77: No need to use `ToLowerInvariant` since the comparison are later done using `OrdinalIgnoreCase`. That would remove this allocation, and maybe make the whole thing alloc-free.

PhyxionNL: Correct, this was a left over from previous version, removed 👍

src/ImageSharp.Web/FormatUtilities.cs
```diff
@@ -19,8 +19,7 @@ namespace SixLabors.ImageSharp.Web
     /// </summary>
     public sealed class FormatUtilities
     {
-        private readonly IImageFormat[] imageFormats;
-        private readonly Dictionary<IImageFormat, string[]> fileExtensions = new Dictionary<IImageFormat, string[]>();
+        private readonly List<string> fileExtensions = new List<string>();
         private readonly Dictionary<string, string> fileExtension = new Dictionary<string, string>();
```
sebastienros, line 23: Should it be renamed `extensionsByMimeType` to make the code clearer? It's hard to understand what it is otherwise, and the two collections are very close.

src/ImageSharp.Web/FormatUtilities.cs
JimBobSquarePants: They'll get loaded in the order as defined in the `Configuration` class. Current order is png, jpg, gif, bmp, tga. Someone can define their own configuration instance via `ImageSharpMiddlewareOptions` if need be. 

Can't believe I didn't know about the `ROS<char>` equality comparison. Super useful!

src/ImageSharp.Web/FormatUtilities.cs
JimBobSquarePants: Agreed. My original naming is bad. We can do a bit of cleanup here actually. No need to store the `imageFormats` array now and should use C# 9 initializer for the fields to make things more terse.

src/ImageSharp.Web/FormatUtilities.cs
PhyxionNL: I think it would probably be the best to keep order the same as the one in the config, otherwise it would be another thing to configure. The order of png, jpg seems like the good default based on web usage. With webp maybe it can be added before png (or after).

src/ImageSharp.Web/FormatUtilities.cs
PhyxionNL: Updated.

sebastienros (approved):

src/ImageSharp.Web/FormatUtilities.cs
JimBobSquarePants: I'm actually gonna PR upstream into the 2.0 alphas to change the registration order to match most common usage.

JimBobSquarePants:
This was fun! Thanks both of you!

PR #189, opened 2021-11-26
LogServedAction+ changed to Debug rather than Info

marklonquist:
The logging (especially for the "LogServedAction" is kinda excessive at the moment. I suggest lowering the level to Debug so it doesn't log so much.

### DIFF
--- a/src/ImageSharp.Web/Middleware/LoggerExtensions.cs
+++ b/src/ImageSharp.Web/Middleware/LoggerExtensions.cs
@@ -43,7 +43,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
                 formatString: "The image '{Uri}' was not modified");
 
             LogPreconditionFailedAction = LoggerMessage.Define<string>(
-                logLevel: LogLevel.Debug,
+                logLevel: LogLevel.Warning,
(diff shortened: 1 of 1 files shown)


marklonquist:
See https://github.com/SixLabors/ImageSharp.Web/issues/190 by my colleague

deanmarcussen:
The last time we reviewed this, it was suggested we follow what the ASP.NET `StaticFileMiddleware` was doing, in the case of file served, I believe they are using `Information`

https://github.com/dotnet/aspnetcore/blob/e1a984898eb5b209975185020c4b1cb7990bcd50/src/Middleware/StaticFiles/src/LoggerExtensions.cs#L35

For info here is the pr https://github.com/SixLabors/ImageSharp.Web/issues/143

I see in this one, you're actually moving levels up, that shouldn't go up, and maybe moving levels down that are maybe ok where they are.

`Information` is quite a noisy log level to use in general, perhaps you need some log filtering on levels, rather than reducing the log level?

marklonquist:
ah I see .. alright - I'll close this.. Will set to warning for "SixLabors.ImageSharp.Web". Thanks.

PR #194, opened 2021-12-10
Restore ability to use null cache

kroymann:
### Prerequisites

### Description
Per the discussion in #193, this PR restores the ability to effectively disable the caching logic through the use of a "NullCache".  The most important part of this change is restoring the ability to stream the response directly from the processed image output, which is how it worked before v1.0.3.  In v1.0.3, the `WriterWorkers` pattern was introduced, and a side effect of that change was that the response was now _always_ streamed from the cache, which broke the ability to use a NullCache.  To address this, I removed the `ReadWorkers`/`WriterWorkers` stuff and replaced it with a more standard reader/writer locking pattern using a synchronization library ported from my company's codebase.  This made it so that the processed image stream could remain available beyond the section of code protected inside the writer lock, and thus be available for use in generating the response.

A key sticking point here (obviously) is benchmark testing this to see if performance was improved or degraded by this change.

### DIFF
--- a/src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
+++ b/src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
@@ -227,10 +227,13 @@ namespace SixLabors.ImageSharp.Web.Synchronization
             }
 
             public bool Equals(
-#if NETCOREAPP3_1_OR_GREATER
-                [System.Diagnostics.CodeAnalysis.AllowNull]
-#endif
+#if NET5_0_OR_GREATER
+                RefCountedValue? other)
+#elif NETCOREAPP3_1_OR_GREATER
+                [System.Diagnostics.CodeAnalysis.AllowNull] RefCountedValue other)
+#else
                 RefCountedValue other)
+#endif
                 => (other != null) && (this.RefCount == other.RefCount) && EqualityComparer<TValue>.Default.Equals(this.Value, other.Value);
 
             public override bool Equals(object? obj)
--- a/src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
+++ b/src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
@@ -1,0 +1,193 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+#nullable enable
+
+namespace SixLabors.ImageSharp.Web.Synchronization
+{
+    /// <summary>
+    /// Represents a thread-safe collection of reference-counted key/value pairs that can be accessed by multiple
+    /// threads concurrently. Values that don't yet exist are automatically created using a caller supplied
+    /// value factory method, and when their final refcount is released they are removed from the dictionary.
+    /// </summary>
+    /// <typeparam name="TKey">The type of the key.</typeparam>
+    /// <typeparam name="TValue">The value for the dictionary.</typeparam>
+    public class RefCountedConcurrentDictionary<TKey, TValue>
+        where TKey : notnull
+        where TValue : class
+    {
+        private readonly ConcurrentDictionary<TKey, RefCountedValue> dictionary;
+        private readonly Func<TKey, TValue> valueFactory;
+        private readonly Action<TValue>? valueReleaser;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RefCountedConcurrentDictionary{TKey, TValue}"/> class that is empty,
+        /// has the default concurrency level, has the default initial capacity, and uses the default comparer for the key type.
+        /// </summary>
+        /// <param name="valueFactory">Factory method that generates a new <typeparamref name="TValue"/> for a given <typeparamref name="TKey"/>.</param>
+        /// <param name="valueReleaser">Optional callback that is used to cleanup <typeparamref name="TValue"/>s after their final ref count is released.</param>
+        public RefCountedConcurrentDictionary(Func<TKey, TValue> valueFactory, Action<TValue>? valueReleaser = null)
+            : this(new ConcurrentDictionary<TKey, RefCountedValue>(), valueFactory, valueReleaser)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="RefCountedConcurrentDictionary{TKey, TValue}"/> class that is empty,
+        /// has the default concurrency level and capacity,, and uses the specified  <see cref="IEqualityComparer{TKey}"/>.
+        /// </summary>
+        /// <param name="comparer">The <see cref="IEqualityComparer{TKey}"/> implementation to use when comparing keys.</param>
+        /// <param name="valueFactory">Factory method that generates a new <typeparamref name="TValue"/> for a given <typeparamref name="TKey"/>.</param>
+        /// <param name="valueReleaser">Optional callback that is used to cleanup <typeparamref name="TValue"/>s after their final ref count is released.</param>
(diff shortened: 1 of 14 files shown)


JimBobSquarePants:
@kroymann I'm struggling to find the time to benchmark this so if you have the opportunity please do.

deanmarcussen:
I had a quick look this morning.

The results are both very promising, and suspicously close.

It either indicates that the LRU cache is blocking any meaningful test of this piece of functionality (I don't think so, but need to have a longer look at code to say for sure), or both locking arrangements are having no meaningful impact on performance, and we've at the limit of what we can serve (i.e. limited by disk speed, or the framework itself).

I'm tending to the second, as the most meaningful change I could make to impact rps was to set the samples project logging to "Warning" instead of "Debug". (basically drop all the expensive logging)

Need to find some time to look at the code changes, and do another run with the samples project updated to .NET 6.

JimBobSquarePants:
Thanks @deanmarcussen 

I think it might be an idea to try both this and master with a much reduced timespan for LRU cache. It's sitting at 5 minutes just now which will be interfering with results. It would be amazing if you could document the setup you use to run the benchmark as I had a go at reading the Crank docs again and felt a bit overwhelmed. 

A .NET 6 run would definitely be a good idea.

src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
```diff
@@ -0,0 +1,243 @@
...
+            public readonly int RefCount;
+#pragma warning restore SA1401 // Fields should be private
+
+            public RefCountedValue(TValue value, int refCount)
+            {
+                this.Value = value;
+                this.RefCount = refCount;
+            }
+
+            public bool Equals(
```
sebastienros, line 229: I am getting ` error CS8767: Nullability of reference types in type of parameter 'other' of 'bool RefCountedValue.Equals(RefCountedValue other)' doesn't match implicitly implemented member 'bool IEquatable<RefCountedValue>.Equals(RefCountedValue other)' (possibly because of nullability attributes)` when running the benchmarks

kroymann: Weird.  This compiles cleanly in both netcoreapp2.1 and 3.1 for me, and I'm able to run the benchmarks in both TFMs as well....?  @sebastienros Anything special about how you're trying to run the benchmarks?

sebastienros: To be more precise, using crank, doing a web load benchmark independent from BDN. It works fine on master though. Full stack:

```
Command:
dotnet publish ImageSharp.Web.Sample.csproj -c Release -o /tmp/benchmarks-agent/benchmarks-server-1/2ns2okc1.ov0/ImageSharp.Web/samples/ImageSharp.Web.Sample/published /p:MicrosoftNETCoreAppPackageVersion=3.1.21 /p:MicrosoftAspNetCoreAppPackageVersion=3.1.21 /p:MicrosoftNETCoreApp31PackageVersion=3.1.21 /p:MicrosoftNETPlatformLibrary=Microsoft.NETCore.App /p:RestoreNoCache=true --framework netcoreapp3.1 --self-contained -r linux-x64 
Microsoft (R) Build Engine version 16.7.2+b60ddb6f4 for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored /tmp/benchmarks-agent/benchmarks-server-1/2ns2okc1.ov0/ImageSharp.Web/src/ImageSharp.Web/ImageSharp.Web.csproj (in 297 ms).
  Restored /tmp/benchmarks-agent/benchmarks-server-1/2ns2okc1.ov0/ImageSharp.Web/samples/ImageSharp.Web.Sample/ImageSharp.Web.Sample.csproj (in 297 ms).
Synchronization/RefCountedConcurrentDictionary.cs(229,25): error CS8767: Nullability of reference types in type of parameter 'other' of 'bool RefCountedValue.Equals(RefCountedValue other)' doesn't match implicitly implemented member 'bool IEquatable<RefCountedValue>.Equals(RefCountedValue other)' (possibly because of nullability attributes). [/tmp/benchmarks-agent/benchmarks-server-1/2ns2okc1.ov0/ImageSharp.Web/src/ImageSharp.Web/ImageSharp.Web.csproj]
Exit code: 1
```

kroymann: I just pushed a small change to this code that adjusts the nullability attributes when compiling with net5.0 or higher.  This is preemptively getting ahead of any update to this codebase to target net6.0, and _maybe_ it will address whatever issue you hit running benchmarks?

src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
```diff
@@ -0,0 +1,246 @@
...
+
+            public bool MoveNext() => this.enumerator.MoveNext();
+
+            public void Reset() => this.enumerator.Reset();
+        }
+
+        /// <summary>
+        /// Simple immutable tuple that combines a <typeparamref name="TValue"/> instance with a ref count integer.
+        /// </summary>
+        private class RefCountedValue : IEquatable<RefCountedValue>
```
sebastienros, line 216: Can you use struct records here instead ?

src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
sebastienros: With your changes it builds on net6.0. At least it's unblocking me.

src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
kroymann: I benchmarked this using both class and struct for this type and determined that using a class executes more quickly and allocates less memory.  I believe this happens because `ConcurrentDictionary` can use an optimized code path that leverages atomic writes when TValue is a class, but has to fall back on a less efficient path that allocates when TValue is a struct.
```
|                          Method |      Mean |    Error |   StdDev |  Gen 0 | Allocated |
|-------------------------------- |----------:|---------:|---------:|-------:|----------:|
|       Class_GetAndReleaseNewKey | 129.92 ns | 0.554 ns | 0.518 ns | 0.0095 |      80 B |
|      Struct_GetAndReleaseNewKey | 147.59 ns | 0.687 ns | 0.573 ns | 0.0067 |      56 B |
|  Class_GetAndReleaseExistingKey | 142.32 ns | 1.159 ns | 1.027 ns | 0.0076 |      64 B |
| Struct_GetAndReleaseExistingKey | 177.69 ns | 0.813 ns | 0.721 ns | 0.0134 |     112 B |
|            Class_GetExistingKey |  69.75 ns | 0.301 ns | 0.267 ns | 0.0038 |      32 B |
|           Struct_GetExistingKey |  89.81 ns | 0.682 ns | 0.638 ns | 0.0067 |      56 B |
```

I have not yet experimented with using record types (in part because this codebase is still targeting netcoreapp2.1 and 3.1).

JimBobSquarePants: Let's leave changing target frameworks to V2. I don't want your hard work delayed by our upstream work.

JimBobSquarePants:
@sebastienros Thanks for the numbers. I can see a bit of give and take but otherwise fairly even. For the additional feature I think the change is worth it.

src/ImageSharp.Web/Synchronization/RefCountedConcurrentDictionary.cs
sebastienros: If not a struct record, tuples would still be good while keeping the code simpler. But I don't know about tfm requirements either, so you'll decide.

JimBobSquarePants (approved):
Let's get this merged in. Perf is comparable and we can always iterate further.

PR #203, opened 2022-01-21
Replace generic dictionary for commands with specialized collection type.

JimBobSquarePants:
### Prerequisites

### Description

This breaking change is required in order to allow users to easily insert commands  into the processing pipeline via the `OnParseCommandsAsync`. 

It introduces a new type `CommandCollection` based upon [ ` KeyedCollection<string, KeyValuePair<string, string>>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.keyedcollection-2?view=net-6.0) to replace the generic `IDictionary<string, string>` previously used to house parsed commands.

The requirement was brought to my attention in #191 where an issue was correctly identified when attempting to insert commands into the beginning of the command collection. 

While the change is breaking and changes the signature of a fundamental component (`IImageWebProcessor`) I believe that it is an important change to make and we should do so now before it is too late. 

Following the change, automatically inserting the command for a potential `AutoOrientWebProcessor` processor to manage EXIF rotation would be as simple as following.

```c#
c.Commands.Insert(0, new(AutoOrientWebProcessor.Key, null));
```

CC @ronaldbarendse  as I want to know what affect this would have on Umbraco.
CC @deanmarcussen as I want to know what affect this would have on Orchard.


deanmarcussen:
It will be marginally breaking for us, only in the sense that the api has changed, so we can't easily release to our preview feed, and get some people trying the new version, as we don't like to push anything that isn't on NuGet.

We do quite a lot with commands, for DDOS prevention, and such, but it's mostly about removal. Either way the api still does what it used to, so all good from us.

JimBobSquarePants:
@deanmarcussen Thanks for getting back to me. Good to hear it's not an issue.

ronaldbarendse:
Thanks for bringing this to my attention 👍🏻 We'll only update the ImageSharp dependency to v2 in the next major Umbraco version (v10, planned Q2, 2022), so breaking this API shouldn't be a big issue for us.

However, I'm not aware of any other breaking API changes in ImageSharp/ImageSharp.Web and this would prevent users from upgrading Umbraco 9 sites to use v2, as the codebase uses the affected `OnParseCommandsAsync` and adds a `CropWebProcessor` (a custom `IImageWebProcessor`). It would be a shame if users can't add WebP support and improved memory management, either because Umbraco 10 isn't released yet or they can't upgrade their site (e.g. because of other dependencies).

I've quickly tested this on an Umbraco 9.2.0 project using the nightly NuGet feed (SixLabors.ImageSharp 2.0.0-alpha.0.159 and SixLabors.ImageSharp.Web 2.0.0-alpha.0.3) and both the command parsing (configurable max width/height) and the custom crops still work.

This is the only concern I currently have with these changes, but I also see the value in allowing commands to be inserted (adding auto-rotation, as you already showed).

src/ImageSharp.Web/Commands/CommandCollection.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="CommandCollection"/> class.
+        /// </summary>
+        public CommandCollection()
+            : this(StringComparer.OrdinalIgnoreCase)
+        {
+        }
+
+        private CommandCollection(IEqualityComparer<string> comparer)
+            : base(comparer)
```
ronaldbarendse, line 26: Keep in mind that this constructor already creates an internal lookup dictionary: https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.keyedcollection-2.-ctor?view=net-6.0#system-collections-objectmodel-keyedcollection-2-ctor(system-collections-generic-iequalitycomparer((-0))-system-int32).

ronaldbarendse (changes requested):
A lot of the breaking changes can be avoided by having `CommandCollection` implement `IDictionary<string, string>` and forward that to the `KeyedCollection<>`, right?

That would make accessing the `context.Commands` in the `OnParseCommandsAsync` work, as it can still use the members of the `IDictionary<string, string>` on the new collection. And `IImageWebProcessor` can keep the existing signature, as that doesn't alter the order of the command and thus doesn't need the concrete implementation.

src/ImageSharp.Web/Commands/CommandCollection.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+        }
+
+        /// <summary>
+        /// Gets an <see cref="ICollection{T}"/> representing the keys of the collection.
+        /// </summary>
+        public ICollection<string> Keys => this.keys;
```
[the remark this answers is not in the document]
JimBobSquarePants: The internal lookup `Dictionary.Keys` property isn't actually in order.

src/ImageSharp.Web/Commands/CommandCollection.cs
JimBobSquarePants: I considered implementing `IDictionary<string, string>` but it leads to a confusing API since the type inherits `Collection<T>`. `Contains` vs `ContainsKey` for example. There's also virtualization issues passing around _IDictionary<T>_ all the time. 

I'm not quite sure what you're pointing out re the constructor. Are you suggesting using the internal dictionary to implement the interface or for something else?

src/ImageSharp.Web/Commands/CommandCollection.cs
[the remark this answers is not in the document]
ronaldbarendse: Ah, of course, I already thought I was missing something! 😖

I see this is only used in `ImageSharpMiddleware.StripUnknownCommands()`, so might be worthwhile to convert this into something like this:

```suggestion
        public IEnumerable<string> Keys
        {
            get
            {
                foreach (var item in this)
                {
                    yield return this.GetKeyForItem(item);
                }
            }
        }
```

src/ImageSharp.Web/Commands/CommandCollection.cs
ronaldbarendse: Confusion can be avoided by using an explicit interface implementation (which might already be required for some conflicting members)...

The point I wanted to make, is that this implementation stores the items both in the collection and internal lookup dictionary, in addition to keeping a sorted list of keys.

Too bad `OrderedDictionary` isn't using generics, although I see there is an [internal MS implementation](https://referencesource.microsoft.com/#System.ServiceModel.Internals/System/Runtime/Collections/OrderedDictionary.cs).

skttl:
> However, I'm not aware of any other breaking API changes in ImageSharp/ImageSharp.Web and this would prevent users from upgrading Umbraco 9 sites to use v2, as the codebase uses the affected `OnParseCommandsAsync` and adds a `CropWebProcessor` (a custom `IImageWebProcessor`). It would be a shame if users can't add WebP support and improved memory management, either because Umbraco 10 isn't released yet or they can't upgrade their site (e.g. because of other dependencies).

I thought it was possible to swap out the implementation in Umbraco, so potentially a package could bridge Umbraco 9 to ImageSharp 2?

src/ImageSharp.Web/Commands/CommandCollection.cs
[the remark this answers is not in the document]
JimBobSquarePants: That's a neat suggestion! The enumerator should still be fast here.

I'll not commit this directly as it means I can remove a bunch of overrides at the same time.

ronaldbarendse:
> I thought it was possible to swap out the implementation in Umbraco, so potentially a package could bridge Umbraco 9 to ImageSharp 2?

@skttl Umbraco does indeed use abstractions (for image URL generation and getting image dimensions), but `Umbraco.Cms.Infrastructure` and `Umbraco.Cms.Web.Common` still have a tight integration with `ImageSharp` and `ImageSharp.Web`, respectively. This means you can't install a version that isn't API compatible (as that would throw a `MissingMethodException`), although you can use a different package/service.

@JimBobSquarePants I've done some light hacking to get this PR compatible with Umbraco 9: https://github.com/ronaldbarendse/ImageSharp.Web/commit/66f843d3e65e5abf1963b5b4d6654de86cc8cb98. Having `CommandCollection` implement `IDictionary<string, string>` was quite trivial and removes the need to change the `IImageWebProcessor.Process()` signature. The only thing that's a bit nasty, is the additional `ImageCommandContext.OrderedCommands` property, so the existing `Commands` property can be kept unchanged as well. Let me know what you think of this!

JimBobSquarePants:
@ronaldbarendse I hard a look at your branch and while the refactor is impressive I still believe the dedicated type is the right way to go. 

- Exposing `IDictionary` in this manner exposes the less efficient Keys/Values properties where we allocate a new list each time each of these properties are called. If people call them in from an `IImageWebProcessor` that's gonna suck. We are also now stuck again with the virtual dispatch performance issues. 
- The obsolete code that you've already identified as nasty 😄 is most definitely that. I wouldn't introduce an obsolete attribute for a major version. 

Oh and re upgrading a major... There's [precedence](https://github.com/umbraco/Umbraco-CMS/pull/9920/files) - Maybe we can convince the powers that be to let it happen. The memory usage improvements alone are well worth it and there's not many libraries that unit test as much as we do.

src/ImageSharp.Web/Commands/CommandCollectionExtensions.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public static class CommandCollectionExtensions
+    {
+        /// <summary>
+        /// Gets the value associated with the specified key or the default value.
+        /// </summary>
+        /// <param name="collection">The collection instance.</param>
+        /// <param name="key">The key of the value to get.</param>
+        /// <returns>The value associated with the specified key or the default value.</returns>
+        public static string GetValueOrDefault(this CommandCollection collection, string key)
```
deanmarcussen, line 19: I didn't see this being used anywhere

Do you really want it? Wouldn't just using `TryGetValue`  be enough / better?

src/ImageSharp.Web/Commands/PresetOnlyQueryCollectionRequestParser.cs
```diff
@@ -31,31 +31,33 @@ public class PresetOnlyQueryCollectionRequestParser : IRequestParser
             this.presets = ParsePresets(presetOptions.Value.Presets);
 
         /// <inheritdoc/>
-        public IDictionary<string, string> ParseRequestCommands(HttpContext context)
+        public CommandCollection ParseRequestCommands(HttpContext context)
         {
             if (context.Request.Query.Count == 0 || !context.Request.Query.ContainsKey(QueryKey))
             {
-                return new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                return new();
```
deanmarcussen, line 38: Can you use a static instance here? Maybe better for no-ops

src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
```diff
@@ -15,18 +14,20 @@ namespace SixLabors.ImageSharp.Web.Commands
     public sealed class QueryCollectionRequestParser : IRequestParser
     {
         /// <inheritdoc/>
-        public IDictionary<string, string> ParseRequestCommands(HttpContext context)
+        public CommandCollection ParseRequestCommands(HttpContext context)
         {
             if (context.Request.Query.Count == 0)
             {
-                return new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+                return new();
```
deanmarcussen, line 21: and a static here

src/ImageSharp.Web/Commands/CommandCollectionExtensions.cs
JimBobSquarePants: It’s used a lot in the individual processors. Allows much less code when passing values to the command parser which already handles the default instance

src/ImageSharp.Web/Commands/PresetOnlyQueryCollectionRequestParser.cs
JimBobSquarePants: I’ll have a look. Should be possible

ronaldbarendse:
@JimBobSquarePants My attempt was mainly to see whether it was at least possible to minimize the breaking changes so it would still work on an existing Umbraco 9 project. You make valid points about the property/virtual dispatch performance issues and the nastiness it requires, so lets break some (more) stuff! To fix things and make it better of course! 😉 

Regarding the precedence though: that major version bump in `LightInject` got reverted, same as with bumping the `NPoco` dependency... Umbraco 10 will be an LTS release, so I'd rather see a clean and performing v2 dependency included in the end. Lets have a chat if we need to align our release plans to make that possible.

--------------

I do have some suggestions on adding a couple of convenience methods to the `CommandCollection` (not sure if they all make sense, but I'll leave you to be the judge of that):

```c#
public void Add(string key, string value) => this.Add(new(key, value));

public void Insert(int index, string key, string value) => this.Insert(index, new(key, value));

public bool TryGetValue(string key, out string value)
{
    if (this.TryGetValue(key, out KeyValuePair<string, string> keyValue))
    {
        value = keyValue.Value;
        return true;
    }

    value = default;
    return false;
}
```

src/ImageSharp.Web/Commands/CommandCollectionExtensions.cs
ronaldbarendse: Don't forget to remove the `DictionaryExtensions` class if that becomes obsolete by this change!

And if my suggestion regarding adding a `TryGetValue(string key, out string value)` overload doesn't get included, this makes even more sense, as you don't need the `KeyValuePair<string, string>` if you already know the key...

src/ImageSharp.Web/Commands/PresetOnlyQueryCollectionRequestParser.cs
ronaldbarendse: Would make sense to add it as `CommandCollection.Empty` 👍🏻

JimBobSquarePants: Actually this is correct as it is. The parser returns an empty collection that can then be augmented via `OnParseCommandsAsync` we don't want a shared instance for that.

src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
JimBobSquarePants: Same as `PresetOnlyQueryCollectionRequestParser` we need to be able to augment the collection.

deanmarcussen: Yeah you're totally right, and I knew that a year ago when I last worked on this bit of code.

If you feel like popping a comment in to remind me next year, go for it!

JimBobSquarePants:
Just added the comments for future us. Once this builds I think we're ready to merge.

deanmarcussen (approved):
Side note, not sure if you're planning on an RC of v2 for `ImageSharp` and `ImageSharp.Web`, to NuGet, but if you are we'll take it in OC straight away. 

Plenty of people real keen for webp, and I'm looking forward to the new memory allocator

PR #206, opened 2022-01-27
Abstract cache key creation to ICacheKey

ronaldbarendse:
### Prerequisites

### Description
This implements the feature discussed in https://github.com/SixLabors/ImageSharp.Web/discussions/145 and is related to PR https://github.com/SixLabors/ImageSharp.Web/pull/147 that uses a setting to ignore the hostname.

To summarize: both `IRequestParser` and `IImageProvider` can return different source/processed images based on whatever they extract from the `HttpContext`, but the `ImageSharpMiddleware` always caches the image based on the lowercased request host, path and parsed commands.

This leads to creating cache keys that are either:
1. Too unique, e.g. when you have multiple hostnames configured (even if it's just the `www.` prefix and domain root). This will generate and cache 2 images for `example.com/imagesharp-logo.png` and `www.example.com/imagesharp-logo.png`, which is a waste of processing (CPU/memory) and disk resources (there's also no cache trimming, so these won't get automatically cleaned up!).
2. Not unique enough, e.g. when you use request headers, user authentication or even a query string (as only known commands are used in the cache key) to return a different source image.

The first commit in this PR simply moves the existing code to the default `CacheKey` implementation. The second one adds 4 alternative implementations that either use the absolute or relative URI and optionally lowercases that value. Maybe it's better to move these out to a test project or the documentation, but I wanted to share how easy implementing this new interface can be!


src/ImageSharp.Web/Caching/CacheKey.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+using Microsoft.AspNetCore.Http;
+using SixLabors.ImageSharp.Web.Commands;
+
+namespace SixLabors.ImageSharp.Web.Caching
+{
+    /// <summary>
+    /// Creates a cache key based on the lowercased request host, path and commands.
+    /// </summary>
+    public class CacheKey : ICacheKey
```
JimBobSquarePants, line 13: I think we should only ship with the two implementations as described in https://github.com/SixLabors/ImageSharp.Web/pull/147#issuecomment-1023047344 although I like your naming convention better.

- `UriAbsoluteCacheKey`
- `UriRelativeCacheKey`

I can't see a scenario where we would want a case sensitive cache key e.g `UriAbsoluteCacheKey` as that opens us up to potential DOS so we should always convert the string to lower invariant.

From a performance perspective we should use `string.Create` in conjunction with `Span<char>.ToLowerInvarient()` to ensure that string creation is snappy. [`Urihelper`](https://source.dot.net/#Microsoft.AspNetCore.Http.Extensions/UriHelper.cs,e32d072f616dc38d) has optimized code you can use as a reference but we'd have to change a few things to do the required casing work.

ronaldbarendse: Depending on the OS (or file system configuration: https://docs.microsoft.com/en-us/windows/wsl/case-sensitivity#modify-case-sensitivity), the path can be case sensitive, which means different casing can resolve to different files and also that wrong casing wouldn't find the original image in the first place... So the cache hash should only be 'normalized' when the file system is case insensitive. I'm not sure if this can be easily determined at runtime, so adding a setting for this would be sensible.

The parsed commands are already case insensitive (or at least the keys), because the default implementation uses an `InvariantIgnoreCase` comparer (and gets them from the query string, which isn't case sensitive according to the RFCs). Although it ultimately depends on the processor, command values should also be case insensitive (enum parsing already ignores casing). It would be good to ensure this value is always lowercased, so changes in the command casing won't cause different cache hashes.

JimBobSquarePants: Case sensitivity of file requests is dealt with via individual  provider implementations. They don’t use the key to return the source image, they use the input uri. 

Only the cache uses the key and it’s imperative that the generated hash is case insensitive.

JimBobSquarePants: Yep. Case sensitivity of file requests is dealt with via individual  provider implementations. They don’t use the key to return the source image, they use the input uri. If nothing is found, nothing is cached.

Rethinking about this then you're correct but the implementations should be non inheriting since we don't want a virtual method as a default implementation on a fast path. We're already paying a virtualization cost via interfaces.

So, as you say, to cover our backs, your implementations would be:

- `UriAbsoluteCacheKey`
- `UriRelativeCacheKey`
- `UriAbsoluteCaseInsensitiveCacheKey`
- `UriRelativeCaseInsensitiveCacheKey`

You're gonna want custom implementations of the `UriHelper` methods for handling the concatenation + casing to avoid duplicate string generation for all scenarios since the caching of commands should, as you've rightly identified, be case insensitive. A processor not using the case insensitive `CommandCollection` for command key matching would be considered a bug in that implementation.

ronaldbarendse: That's exactly my point: the `IImageProvider` implementations correctly deal with the case sensitivity, but the cache key doesn't. So if you have multiple files that only differ in casing, it will only cache 1 generated image (with the metadata next to it), depending on which was requested first, the TLRU cache, file write times and other weird factors:

Different cased files

You can test this yourself by running the sample site in WSL and creating the copies using `cp imagesharp-logo.png ImageSharp-logo.png`.

Having a setting to disable lowercasing the cache key would at least make it easy to switch (if you're running into this edge case), but also ensure custom implementations of `ICacheKey` are exposed to this setting if it's passed into the `Create` method. And it requires less classes/implementations 😄

JimBobSquarePants: I wouldn't add an option for this for two reasons. 

1. Each implementation would have to respect that boolean option (complexity).
2. Each implementation would have to parse that boolean per request (performance). 

Making the case sensitivity a detail of the registered implementation provides us with the most flexibility. Registering an `ICacheKey` implementation is no more difficult than setting an option.

src/ImageSharp.Web/Caching/ICacheKey.cs
```diff
@@ -0,0 +1,25 @@
...
+        /// <summary>
+        /// Creates the cache key based on the specified context and commands.
+        /// </summary>
+        /// <param name="context">The HTTP context.</param>
+        /// <param name="commands">The commands.</param>
+        /// <param name="caseSensitive">If set to <c>true</c> the cache key should be generated in a case sensitive manner.</param>
+        /// <returns>
+        /// The cache key.
+        /// </returns>
+        string Create(HttpContext context, CommandCollection commands, bool caseSensitive);
```
sebastienros, line 23: I think `caseSensitive` should be an implementation detail of the concrete class, or an option at this level.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -96,6 +96,33 @@ public static IImageSharpBuilder SetCache(this IImageSharpBuilder builder, Func<
...
+        /// <summary>
+        /// Sets the given <see cref="ICacheKey"/> adding it to the service collection.
+        /// </summary>
+        /// <typeparam name="T">The type of class implementing <see cref="ICacheKey"/> to add.</typeparam>
+        /// <param name="builder">The core builder.</param>
+        /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
+        public static IImageSharpBuilder SetCacheKey<T>(this IImageSharpBuilder builder)
+            where T : class, ICacheKey
+        {
+            var descriptor = new ServiceDescriptor(typeof(ICacheKey), typeof(T), ServiceLifetime.Singleton);
```
sebastienros, line 108: Looks like the default implementations of `ICacheKey` don't have any state. In that case it's better to use `Transient` since it eliminates a lookup in the DI entries. It will just emit an efficient `new`.

JimBobSquarePants: That's interesting. I always thought that would be the opposite. That the lookup would be as expensive for both and that since it was stateless we can use the singleton and avoid the per request allocation.

sebastienros: I'll ask a few gods to confirm.

halter73: It depends on how efficient allocating a new instance of T each time is vs retrieving the same instance.

When you're calling `sp.GetService(typeof(ICacheKey))`, the provider will have to search [a ConcurrentDictionary](https://github.com/dotnet/runtime/blob/479ee518f32cfaf53285b915265ccead9b6b13c9/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/CallSiteFactory.cs#L18) for an `ICacheKey` service "call site" regardless. This happens for any service type no matter the lifetime when resolved directly with `GetService`.

A transient lifetime means there's not an additional lookup into [another ConcurrentDictionary](https://github.com/dotnet/runtime/blob/479ee518f32cfaf53285b915265ccead9b6b13c9/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/ServiceProviderEngineScope.cs) to try to get an already-resolved service. Singleton services are simply resolved from a special "root scope", so they're not all that different from scoped services as far as this lookup goes.

If you have a service with a constructor like `ServiceA(ICacheKey)` and you call `sp.GetService(typeof(ServiceA))`, then the service provider will lookup `ServiceA` from the call site dictionary, but it will not need to lookup the `ICacheKey` service from the same dictionary. `ServiceA`'s dependencies are already embedded in the call site.


If the `ICacheKey` service is transient, `ServiceA` will be constructed by just newing up the ICacheKey implementation type: `new ServiceA(new T())`. If the `ICacheKey` service is singleton or scoped, it will be more like: `new ServiceA(
ResolvedServices.GetOrAdd(_ => new T())`.

The reality is a little more complicated because even with ConcurrentDictionary, there needs to be a lock to ensure the same service doesn't get double initialized withing the same scope. There's also some additional logic to track disposables. You can see the real implementation [here](https://github.com/dotnet/runtime/blob/479ee518f32cfaf53285b915265ccead9b6b13c9/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/ILEmit/ILEmitResolverBuilder.cs#L295-L322). You'll notice that the `else` branch of the `if (callSite.Cache.Location == CallSiteResultCacheLocation.Scope)` is a lot simpler.

ronaldbarendse: I've simply used the same lifetime as in all the other extension methods 😇

src/ImageSharp.Web/Caching/ICacheKey.cs
ronaldbarendse: We're discussing that here: https://github.com/SixLabors/ImageSharp.Web/pull/206#discussion_r794052974.

src/ImageSharp.Web/Caching/CacheKey.cs
ronaldbarendse: The whole point for adding the `bool caseSensitive` parameter is to make sure the implementations _have_ to respect that setting (or at least are directly exposed to it).

The same can be said for the `uint length` parameter currently on `ICacheHash.Create`: all implementations need to ensure they generate a hash of the specified length. Not having a global setting might result in one implementation using a setting to be able to configure it and another always returning a fixed length hash (or just as worse: every implementation using their own setting).

Another very important thing about the `CacheKeyCaseSensitive` setting is that the need for enabling it depends on the OS/platform you're running you application on. You don't want to edit code/create separate builds to make your application work correctly for as case sensitive and insensitive file system 😖

ronaldbarendse: If the setting is really a no-go, I'd suggest to only provide the case insensitive relative (default) and absolute implementations.

JimBobSquarePants: I can smash out the two other implementations for you if you like. They wouldn’t take me long

ronaldbarendse: I'd very much like that, as I'm not yet familiar how to best work with `string.Create` and `Span<char>` 😉

I did create some additional tests and benchmarks, so it should be easy for you to validate the implementations 🎉

ronaldbarendse: I've added the previous cache key creation code as baseline benchmark and the `UriAbsoluteCacheKey` implementation is still faster and allocates less (even though that also includes the request scheme, thus creating a longer cache key):

https://github.com/SixLabors/ImageSharp.Web/blob/8c966ad5c40368b3c561bae7e2783283b37f1a94/tests/ImageSharp.Web.Benchmarks/Caching/CacheKeyBenchmarks.cs#L35-L46

I'm sure you're able to make it even more performant by using `string.Create`/`Span<char>` directly, but the `UriHelper` methods are at least already outperforming the previous `StringBuilder` implementation 👍🏻

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
JimBobSquarePants: Thanks for all the information @halter73 I think in our case any performance difference would be virtually unmeasurable. We'll just stick with the same lifetime for convenience.

tests/ImageSharp.Web.Benchmarks/Caching/CacheKeyBenchmarks.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        |                             Method |     Mean |    Error |   StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
+        |----------------------------------- |---------:|---------:|---------:|------:|--------:|-------:|----------:|
+        |                   LegacyV1CacheKey | 666.8 ns | 13.22 ns | 14.14 ns |  1.00 |    0.00 | 0.1659 |     696 B |
+        |                UriRelativeCacheKey | 358.0 ns |  1.98 ns |  1.65 ns |  0.54 |    0.01 | 0.0706 |     296 B |
+        | UriRelativeCaseInsensitiveCacheKey | 363.1 ns |  7.20 ns | 11.21 ns |  0.55 |    0.02 | 0.0706 |     296 B |
+        |                UriAbsoluteCacheKey | 490.3 ns |  5.14 ns |  4.80 ns |  0.74 |    0.02 | 0.0763 |     320 B |
+        | UriAbsoluteCaseInsensitiveCacheKey | 475.4 ns |  4.18 ns |  3.71 ns |  0.71 |    0.02 | 0.0763 |     320 B |
+        */
+
```
JimBobSquarePants, line 58: @ronaldbarendse  I added the new benchmarks here. Allocations are lower in the new version so would be interested to see what the numbers are on your more powerful computer.

For reference, here's your previous numbers.

|              Method |     Mean |   Error |  StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|-------------------- |---------:|--------:|--------:|------:|--------:|-------:|----------:|
|            Baseline | 497.4 ns | 8.26 ns | 7.33 ns |  1.00 |    0.00 | 0.1106 |     696 B |
| UriRelativeCacheKey | 270.9 ns | 5.47 ns | 7.11 ns |  0.55 |    0.02 | 0.0587 |     368 B |
| UriAbsoluteCacheKey | 447.4 ns | 3.38 ns | 3.16 ns |  0.90 |    0.01 | 0.0939 |     592 B |

JimBobSquarePants:
@ronaldbarendse 

I've written my optimized keygen code and added a legacy version to allow backwards compatibility (_turns out the legacy was adding an extra slash_) think this is good to go. Thanks for your help here, this is a really great addition to the library!

I'll let you have a final read through before merging as I want to make sure I haven't missed anything.

JimBobSquarePants (approved):

src/ImageSharp.Web/Caching/CacheKeyHelper.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        /// <param name="handling">Determines case handling for the result. <paramref name="query"/> is always converted to invariant lowercase.</param>
+        /// <param name="host">The host portion of the uri normally included in the Host header. This may include the port.</param>
+        /// <param name="pathBase">The first portion of the request path associated with application root.</param>
+        /// <param name="path">The portion of the request path that identifies the requested resource.</param>
+        /// <param name="query">The query, if any.</param>
+        /// <returns>The combined URI components, properly encoded for use in HTTP headers.</returns>
+        public static string BuildAbsoluteKey(
+            CaseHandling handling,
+            HostString host,
```
JimBobSquarePants, line 53: I've deliberately chosen to not include the scheme or fragment here.

src/ImageSharp.Web/Caching/UriAbsoluteCacheKey.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using Microsoft.AspNetCore.Http;
+using SixLabors.ImageSharp.Web.Commands;
+
+namespace SixLabors.ImageSharp.Web.Caching
+{
+    /// <summary>
+    /// Creates a cache key based on the request scheme, host, path and commands.
```
ronaldbarendse, line 10:
```suggestion
    /// Creates a cache key based on the request host, path and commands.
```

src/ImageSharp.Web/Caching/UriAbsoluteCaseInsensitiveCacheKey.cs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using Microsoft.AspNetCore.Http;
+using SixLabors.ImageSharp.Web.Commands;
+
+namespace SixLabors.ImageSharp.Web.Caching
+{
+    /// <summary>
+    /// Creates a case insensitive cache key based on the request scheme, host, path and commands.
```
ronaldbarendse, line 10:
```suggestion
    /// Creates a case insensitive cache key based on the request host, path and commands.
```

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -64,6 +64,8 @@ public static IImageSharpBuilder AddImageSharp(this IServiceCollection services)
 
             builder.SetCache<PhysicalFileSystemCache>();
 
+            builder.SetCacheKey<UriRelativeCacheKey>();
```
ronaldbarendse, line 67: Don't we want to use the lowercased cache key by default, as that prevents duplicate cached images and reduces the DOS attack surface?
```suggestion
            builder.SetCacheKey<UriRelativeCaseInsensitiveCacheKey>();
```

This should then also be updated in the sample project `Startup` class...

ronaldbarendse (commented):
I've added some comments: the main thing to currently change, would be the default implementation (that should be the lowercased cache key).

Otherwise, the only remaining thing might be naming: I think we should rename `UriRelativeCaseInsensitiveCacheKey` to `UriRelativeLowerInvariantCacheKey`, as that's what it actually does (case insensitive would be more appropriate when comparing, right?).

JimBobSquarePants:
>Otherwise, the only remaining thing might be naming: I think we should rename UriRelativeCaseInsensitiveCacheKey to UriRelativeLowerInvariantCacheKey, as that's what it actually does (case insensitive would be more appropriate when comparing, right?).

Naming is hard... I chose CaseInsensitive since it described the operational behavior of the key rather than the means that it uses. I'm still torn between the two but leaning towards your convention.

JimBobSquarePants:
I'll fix the build in a mo.

JimBobSquarePants:
OK.... This time we're done.

PR #207, opened 2022-01-27
Rename PhysicalFileSystemProvider to WebRootImageProvider

ronaldbarendse:
### Prerequisites

### Description
As discussed in https://github.com/SixLabors/ImageSharp.Web/discussions/204, this PR renames the existing `PhysicalFileSystemProvider` to `WebRootImageProvider`.

I've also moved the main implementation into a more generic `FileProviderImageProvider` and added the option to specify a path prefix. The `FileProviderImageProvider` can be used when you serve files that can't be resolved from the `WebRootProvider`, e.g. when you've configured a custom `UseStaticFiles()`. In combination with the path prefix, it's possible to move your images outside the webroot:

After `AddImageSharp()`:
```c#
.ClearProviders() // Required because WebRootImageProvider is added by default and matches all requests
.AddProvider(provider =>
{
    var environment = provider.GetRequiredService<IWebHostEnvironment>();
    var root = Path.Combine(environment.ContentRootPath, "Secret");
    var formatUtilities = provider.GetRequiredService<FormatUtilities>();

    return new FileProviderImageProvider(new PhysicalFileProvider(root), formatUtilities, "/secret");
})
.AddProvider<WebRootImageProvider>()
```

And after `UseImageSharp()`:
```c#
app.UseStaticFiles(new StaticFileOptions()
{
    RequestPath = "/secret",
    FileProvider = new PhysicalFileProvider(Path.Combine(env.ContentRootPath, "Secret")),
});
```

Another big benefit of this is the possibility to share the `IFileProvider` implementation between the `StaticFileMiddleware` and `ImageSharpMiddleware`. This could greatly simplify the existing `AzureBlobStorageImageProvider` 👍🏻

### DIFF
--- a/samples/ImageSharp.Web.Sample/Startup.cs
+++ b/samples/ImageSharp.Web.Sample/Startup.cs
@@ -55,7 +55,7 @@ namespace SixLabors.ImageSharp.Web.Sample
                                 provider.GetRequiredService<FormatUtilities>());
                 })
                 .SetCacheHash<CacheHash>()
-                .AddProvider<PhysicalFileSystemProvider>()
+                .AddProvider<WebRootImageProvider>()
                 .AddProcessor<ResizeWebProcessor>()
                 .AddProcessor<FormatWebProcessor>()
                 .AddProcessor<BackgroundColorWebProcessor>()
@@ -131,7 +131,7 @@ namespace SixLabors.ImageSharp.Web.Sample
                 })
                 .SetCacheHash<CacheHash>()
                 .ClearProviders()
-                .AddProvider<PhysicalFileSystemProvider>()
+                .AddProvider<WebRootImageProvider>()
                 .ClearProcessors()
                 .AddProcessor<ResizeWebProcessor>()
                 .AddProcessor<FormatWebProcessor>()
--- a/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
@@ -66,7 +66,7 @@ namespace SixLabors.ImageSharp.Web.DependencyInjection
 
             builder.SetCacheHash<CacheHash>();
 
-            builder.AddProvider<PhysicalFileSystemProvider>();
+            builder.AddProvider<WebRootImageProvider>();
 
             builder.AddProcessor<ResizeWebProcessor>()
                    .AddProcessor<FormatWebProcessor>()
--- a/src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
+++ b/src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
@@ -26,16 +26,28 @@ namespace SixLabors.ImageSharp.Web.Providers
         private readonly FormatUtilities formatUtilities;
 
         /// <summary>
+        /// The path prefix that needs to match the request and is removed before getting the file info.
+        /// </summary>
+        private readonly PathString pathPrefix;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="FileProviderImageProvider"/> class.
         /// </summary>
         /// <param name="fileProvider">The file provider.</param>
         /// <param name="formatUtilities">Contains various format helper methods based on the current configuration.</param>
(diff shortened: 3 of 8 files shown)


JimBobSquarePants:
We don't want to do this as I [discussed](https://github.com/SixLabors/ImageSharp.Web/discussions/204#discussioncomment-2055878) I want to update the provider to match the cache pattern.

ronaldbarendse:
@JimBobSquarePants I would strongly urge you to reconsider this, because the `WebRootProvider` doesn't have to be a physical file system and by using this abstraction, you're also sure that every image that's served by the `StaticFileMiddleware` (requests without processing commands) can also be read and processed by the `ImageSharpMiddleware` (without writing a custom provider).

If you don't like the virtual methods: these can be easily removed together with the `WebRootImageProvider` and replaced with a factory method, while still keeping the extensibility:
```c#
.AddProvider(provider =>
{
    var environment = provider.GetRequiredService<IWebHostEnvironment>();
    var formatUtilities = provider.GetRequiredService<FormatUtilities>();

    return new FileProviderImageProvider(environment.WebRootProvider, formatUtilities);
})
```

Or otherwise just cherry-pick the first commit (https://github.com/SixLabors/ImageSharp.Web/pull/207/commits/a76e58976d6d985cd58c464490bf262d605c375b) and skip the generic `FileProviderImageProvider` and path prefixes 😕

JimBobSquarePants:
I don't want to couple my middleware at any point with the `StaticFileMiddleware`. They should be separate. 

We're not gonna use the `WebRootProvider` We're gonna update the existing provider to use a default root that matches the web provider and use our own `PhysicalFileProvider` based upon that route. That was all previously [discussed](https://github.com/SixLabors/ImageSharp.Web/discussions/204#discussioncomment-2055878) and you're creating unnecessary work for yourself here.

ronaldbarendse:
There's no direct coupling with any other middleware, it's only using the same, ASP.NET Core default, file system access abstractions: https://docs.microsoft.com/en-us/aspnet/core/fundamentals/file-providers.

To illustrate, I've added the `EmbeddedFileProvider` next to (composited with) the default `PhysicalFileProvider` in this PR branch and created a new `embedded.html` demo page in the sample project. This is configured using all default ASP.NET Core file providers:
```c#
env.WebRootFileProvider = new CompositeFileProvider(
    env.WebRootFileProvider,
    new EmbeddedFileProvider(this.GetType().Assembly, "ImageSharp.Web.Sample.Embedded"));
```

This results in `/imagesharp-logo.png` being resolved by the `PhysicalFileProvider` (same as before) and `/sixlabors-logo.png` by the `EmbeddedFileProvider`. This can't be replicated with multiple `IImageProvider`s that use the concrete implementations themselves, as only a single one can match and that would already require a full file exists check...

This pattern is also used by [Razor Class Libraries](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-6.0&tabs=visual-studio) (during development), so static files are served without having to copy them into your project file, see: https://stackoverflow.com/a/58485311/96175.

Not using the `IFileProvider` abstractions would basically mean reinventing the wheel for read-only file access.

JimBobSquarePants:
And that is all fine if someone else wants to do that. However it is out of scope for this project. The whole point of the middleware design is that we provide the very basics and individuals implement custom behaviour they need.

ronaldbarendse:
The [current default `IImageProvider`](https://github.com/SixLabors/ImageSharp.Web/blob/v1.0.5/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs#L69) is [`PhysicalFileSystemProvider`](https://github.com/SixLabors/ImageSharp.Web/blob/v1.0.5/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs), which is already using the `IFileProvider` abstraction! There's nothing in the implementation that's specific to a physical file system though, so what's 'out of scope for this project' about simply renaming this implementation to what it actually does?

I can understand that adding the additional `FileProviderImageProvider` base class and path prefix parsing isn't desirable, which is why I reverted that change... The current PR branch only contains the renamed `IImageProvider` classes and the embedded file demo page in the sample project.

JimBobSquarePants:
We are changing the `PhysicalFileSystemProvider`. We have discussed this ad nauseum and I do not wish to continue this conversation.

PR #208, opened 2022-01-27
Remove usage of IFileProvider in PhysicalFileSystemCache

ronaldbarendse:
### Prerequisites

### Description
As discussed in https://github.com/SixLabors/ImageSharp.Web/discussions/204, the `PhysicalFileSystemCache` doesn't have to use the `IFileProvider` abstraction, as that only provides read-only access and the cache needs write access.

Another benefit of removing the `PhysicalFileProvider` is that we don't have to create the cache directory in the constructor and can remove the dependency on `Microsoft.Extensions.FileProviders.Physical`.

### DIFF
--- a/src/Directory.Build.targets
+++ b/src/Directory.Build.targets
@@ -35,6 +35,5 @@
     <PackageReference Update="Microsoft.AspNetCore.WebUtilities" Version="2.2.0" />
     <PackageReference Update="Microsoft.Extensions.Caching.Abstractions" Version="2.2.0" />
     <PackageReference Update="Microsoft.Extensions.Options.ConfigurationExtensions" Version="2.2.0" />
-    <PackageReference Update="Microsoft.Extensions.FileProviders.Physical" Version="2.2.0"/>
   </ItemGroup>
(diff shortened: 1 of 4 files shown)


JimBobSquarePants:
This is the opposite of what I want to do. [PhysicalFileProvider](https://source.dot.net/#Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs,54) does things for us that we want (path normalization, polling).

ronaldbarendse:
But the `PhysicalFileProvider` is read-only and the `IImageCache` needs write-access to store the processed images, so you still need to do path normalization yourself, right? And polling isn't used and doesn't have a use-case here...

I'm now also completely lost in what pattern you want to introduce, because in PR https://github.com/SixLabors/ImageSharp.Web/pull/207 you're against using the file provider abstraction 😕 Or do you want to instantiate your own `PhysicalFileProvider` based on the `environment.WebRootPath` (and if so, why would that be a better approach then simply using the existing `environment.WebRootFileProvider`)?

PR #212, opened 2022-02-01
Separate cache hash length and folder depth

ronaldbarendse:
### Prerequisites

### Description
As discussed in https://github.com/SixLabors/ImageSharp.Web/discussions/210, this separates the cache hash length and (`PhysicalFileSystemCache` specific) cache folder depth. It also generates shorter cache file names by default by not repeating the characters that are already used in the nested folders.


ronaldbarendse:
I've excluded the LRU cache from code coverage for now: I'm not sure why the coverage decreased, but it's completely missing dedicated tests to begin with.

James-Jackson-South:
@ronaldbarendse Don't worry about the coverage and don't hide it. It's flaky as hell. I had a change recently observed when converting to png's to LFS pointers

JimBobSquarePants:
@ronaldbarendse Don't worry about the coverage and don't hide it. It's flaky as hell. I had a change recently observed when converting to png's to LFS pointers

ronaldbarendse:
Because I think adding a setting just for backwards compatibility isn't the best way forward, I thought about creating a sub-class called `LegacyV1PhysicalFileSystemCache` (similar naming as the `LegacyV1CacheKey`). This could then simply set the base class `useLegacyName` boolean using a protected constructor. But even this introduces code that needs to be cleaned up in the future, so I decided on using the fact that the cache depth and hash length are currently the same to revert back to creating the legacy/longer file paths: https://github.com/SixLabors/ImageSharp.Web/pull/213.

James-Jackson-South:
I think closing this PR and opening another was a bit premature. Best leaving those kinds of decisions to the library owners.

JimBobSquarePants:
I think closing this PR and opening another was a bit premature. Best leaving those kinds of decisions to the library owners.

ronaldbarendse:
Fair enough, although I was pretty sure the other PR that doesn't require additional settings would be preferred 😇

PR #213, opened 2022-02-02
Add CacheFolderDepth option and create shorter file names

ronaldbarendse:
### Prerequisites

### Description
This is my second try to separate the cache hash length and folder depth as discussed in https://github.com/SixLabors/ImageSharp.Web/discussions/210, but without introducing a new option to keep creating backwards compatible cache file paths as done in https://github.com/SixLabors/ImageSharp.Web/pull/212.

In ImageSharp.Web v1 the folder depth is using the same value as the cache hash (with a default of 12), so the nested folders contain all characters from the file name. With this new implementation, it removes the characters used for the folders from the file name (to generate shorter file paths), but this would leave no characters left for the actual file name when these values are equal: `027ee1e815be.meta` would result in `0/2/7/e/e/1/e/8/1/5/b/e/.meta`. In this case, the implementation switches back to the old behavior and generates: `0/2/7/e/e/1/e/8/1/5/b/e/027ee1e815be.meta`. This thus removes the need for an additional setting and reverting back to the legacy behavior is as simple as setting both the `ImageSharpMiddlewareOptions.CacheHashLength` and `PhysicalFileSystemCacheOptions.CacheFolderDepth` to the same value.

### DIFF
--- a/src/ImageSharp.Web/Caching/PhysicalFileSystemCacheOptions.cs
+++ b/src/ImageSharp.Web/Caching/PhysicalFileSystemCacheOptions.cs
@@ -4,7 +4,7 @@
 namespace SixLabors.ImageSharp.Web.Caching
 {
     /// <summary>
-    /// Configuration options for the <see cref="PhysicalFileSystemCache"/>.
+    /// Configuration options for the <see cref="PhysicalFileSystemCache" />.
     /// </summary>
     public class PhysicalFileSystemCacheOptions
     {
@@ -12,6 +12,11 @@ namespace SixLabors.ImageSharp.Web.Caching
         /// Gets or sets the cache folder name.
         /// </summary>
         public string CacheFolder { get; set; } = "is-cache";
+
+        /// <summary>
+        /// Gets or sets the depth of the nested cache folders structure to store the images.
+        /// </summary>
+        public uint CacheFolderDepth { get; set; } = 8;
 
         /// <summary>
         /// Gets or sets the optional cache root folder.
--- a/src/ImageSharp.Web/Caching/PhysicalFileSystemCacheOptions.cs
+++ b/src/ImageSharp.Web/Caching/PhysicalFileSystemCacheOptions.cs
@@ -14,7 +14,7 @@ namespace SixLabors.ImageSharp.Web.Caching
         public string CacheFolder { get; set; } = "is-cache";
 
         /// <summary>
-        /// Gets or sets the depth of the nested cache folders structure to store the images.
+        /// Gets or sets the depth of the nested cache folders structure to store the images. Defaults to 8.
         /// </summary>
         public uint CacheFolderDepth { get; set; } = 8;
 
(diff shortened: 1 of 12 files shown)


tests/ImageSharp.Web.Benchmarks/Caching/StringJoinBenchmarks.cs
```diff
@@ -12,11 +12,13 @@ namespace SixLabors.ImageSharp.Web.Benchmarks.Caching
     public class StringJoinBenchmarks
     {
         private const string Key = "abcdefghijkl";
-        private const int CachedNameLength = 12;
+
+        [Params(0, 8, 12)]
+        public int CacheFolderDepth { get; set; }
```
ronaldbarendse, line 17: Because the generated file paths have different lengths based on the depth, I've tested this with 3 values:

|               Method | CacheFolderDepth |       Mean |     Error |    StdDev | Ratio | RatioSD |  Gen 0 | Allocated |
|--------------------- |----------------- |-----------:|----------:|----------:|------:|--------:|-------:|----------:|
|          String.Join |                0 |  12.047 ns | 0.2181 ns | 0.1933 ns |  1.00 |    0.00 |      - |         - |
| StringBuilder.Append |                0 |  30.396 ns | 0.3118 ns | 0.2764 ns |  2.52 |    0.05 | 0.0229 |     144 B |
|        String.Create |                0 |   2.432 ns | 0.0788 ns | 0.0698 ns |  0.20 |    0.01 | 0.0038 |      24 B |
|                      |                  |            |           |           |       |         |        |           |
|          String.Join |                8 | 200.307 ns | 4.0510 ns | 4.3345 ns |  1.00 |    0.00 | 0.0701 |     440 B |
| StringBuilder.Append |                8 |  61.438 ns | 1.2429 ns | 1.1626 ns |  0.31 |    0.01 | 0.0331 |     208 B |
|        String.Create |                8 |  26.131 ns | 0.2055 ns | 0.1923 ns |  0.13 |    0.00 | 0.0242 |     152 B |
|                      |                  |            |           |           |       |         |        |           |
|          String.Join |               12 | 253.192 ns | 4.2244 ns | 3.9515 ns |  1.00 |    0.00 | 0.0854 |     536 B |
| StringBuilder.Append |               12 |  90.561 ns | 0.9826 ns | 0.9191 ns |  0.36 |    0.01 | 0.0381 |     240 B |
|        String.Create |               12 |  37.491 ns | 0.4276 ns | 0.4000 ns |  0.15 |    0.00 | 0.0293 |     184 B |

JimBobSquarePants (approved):
Kudos. This is a smart workaround. 👍

src/ImageSharp.Web/Caching/PhysicalFileSystemCacheOptions.cs
```diff
@@ -13,6 +13,11 @@ public class PhysicalFileSystemCacheOptions
         /// </summary>
         public string CacheFolder { get; set; } = "is-cache";
 
+        /// <summary>
+        /// Gets or sets the depth of the nested cache folders structure to store the images.
```
JimBobSquarePants, line 17:
```suggestion
        /// Gets or sets the depth of the nested cache folders structure to store the images. Defaults to 8.
```

PR #214, opened 2022-02-03
Decouple PhysicalFileSystemProvider from WebRootFileProvider

JimBobSquarePants:
### Prerequisites

### Description

This PR decouples the `PhysicalFileSystemProvider ` from the environment `WebRootFileProvider` allowing it to work independently bringing greater freedom for source image storage within the physical file system. It should be none breaking since it defaults to the existing source location when the new `PhysicalFileSystemProviderOptions.ProviderRootPath` property is not set.

See https://github.com/SixLabors/ImageSharp.Web/discussions/204#discussioncomment-2055878 for reference. 

I've additionally done a little required cleanup.

### DIFF
--- a/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
+++ b/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
@@ -2,11 +2,13 @@
 // Licensed under the Apache License, Version 2.0.
 
 using System;
+using System.IO;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Hosting;
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Http.Extensions;
 using Microsoft.Extensions.FileProviders;
+using Microsoft.Extensions.Options;
 using SixLabors.ImageSharp.Web.Resolvers;
 
 namespace SixLabors.ImageSharp.Web.Providers
@@ -29,9 +31,11 @@ namespace SixLabors.ImageSharp.Web.Providers
         /// <summary>
         /// Initializes a new instance of the <see cref="PhysicalFileSystemProvider"/> class.
         /// </summary>
+        /// <param name="options">The provider configuration options.</param>
         /// <param name="environment">The environment used by this middleware.</param>
         /// <param name="formatUtilities">Contains various format helper methods based on the current configuration.</param>
         public PhysicalFileSystemProvider(
+            IOptions<PhysicalFileSystemProviderOptions> options,
 #if NETCOREAPP2_1
             IHostingEnvironment environment,
 #else
@@ -40,9 +44,18 @@ namespace SixLabors.ImageSharp.Web.Providers
             FormatUtilities formatUtilities)
         {
             Guard.NotNull(environment, nameof(environment));
-            Guard.NotNull(environment.WebRootFileProvider, nameof(environment.WebRootFileProvider));
 
-            this.fileProvider = environment.WebRootFileProvider;
+            // ContentRootPath is never null.
+            // https://github.com/dotnet/aspnetcore/blob/b89eba6c3cda331ee98063e3c4a04267ec540315/src/Hosting/Hosting/src/WebHostBuilder.cs#L262
+            Guard.NotNullOrWhiteSpace(environment.WebRootPath, nameof(environment.WebRootPath));
+
+            // Allow configuration of the provider without having to register everything
+            PhysicalFileSystemProviderOptions providerOptions = options != null ? options.Value : new();
+            string cacheRootPath = GetProviderRoot(providerOptions, environment.WebRootPath, environment.ContentRootPath);
+
+            // Ensure provider directory is created before initializing the file provider
+            Directory.CreateDirectory(cacheRootPath);
+            this.fileProvider = new PhysicalFileProvider(cacheRootPath);
             this.formatUtilities = formatUtilities;
         }
 
@@ -70,6 +83,24 @@ namespace SixLabors.ImageSharp.Web.Providers
 
             var metadata = new ImageMetadata(fileInfo.LastModified.UtcDateTime, fileInfo.Length);
             return Task.FromResult<IImageResolver>(new PhysicalFileSystemResolver(fileInfo, metadata));
+        }
+
+        /// <summary>
+        /// Determine the provider root path
+        /// </summary>
+        /// <param name="providerOptions">The provider options.</param>
+        /// <param name="webRootPath">The web root path.</param>
+        /// <param name="contentRootPath">The content root path.</param>
+        /// <returns><see cref="string"/> representing the fully qualified provider root path.</returns>
+        internal static string GetProviderRoot(PhysicalFileSystemProviderOptions providerOptions, string webRootPath, string contentRootPath)
+        {
+            string providerRoot = string.IsNullOrWhiteSpace(providerOptions.ProviderRoot)
+                ? webRootPath
+                : providerOptions.ProviderRoot;
+
+            return Path.IsPathFullyQualified(providerRoot)
+                ? providerRoot
+                : Path.GetFullPath(providerRoot, contentRootPath);
         }
     }
 }
--- a/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
+++ b/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
@@ -94,9 +94,9 @@ namespace SixLabors.ImageSharp.Web.Providers
         /// <returns><see cref="string"/> representing the fully qualified provider root path.</returns>
         internal static string GetProviderRoot(PhysicalFileSystemProviderOptions providerOptions, string webRootPath, string contentRootPath)
         {
-            string providerRoot = string.IsNullOrWhiteSpace(providerOptions.ProviderRoot)
+            string providerRoot = string.IsNullOrWhiteSpace(providerOptions.ProviderRootPath)
                 ? webRootPath
-                : providerOptions.ProviderRoot;
+                : providerOptions.ProviderRootPath;
 
             return Path.IsPathFullyQualified(providerRoot)
                 ? providerRoot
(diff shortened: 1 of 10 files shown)


src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -40,9 +44,18 @@ public class PhysicalFileSystemProvider : IImageProvider
...
-            Guard.NotNull(environment.WebRootFileProvider, nameof(environment.WebRootFileProvider));
 
-            this.fileProvider = environment.WebRootFileProvider;
+            // ContentRootPath is never null.
+            // https://github.com/dotnet/aspnetcore/blob/b89eba6c3cda331ee98063e3c4a04267ec540315/src/Hosting/Hosting/src/WebHostBuilder.cs#L262
+            Guard.NotNullOrWhiteSpace(environment.WebRootPath, nameof(environment.WebRootPath));
+
+            // Allow configuration of the provider without having to register everything
+            PhysicalFileSystemProviderOptions providerOptions = options != null ? options.Value : new();
+            string cacheRootPath = GetProviderRoot(providerOptions, environment.WebRootPath, environment.ContentRootPath);
```
ronaldbarendse, line 54: This should be renamed to `providerRootPath`.

deanmarcussen (approved):

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -71,5 +84,23 @@ public Task<IImageResolver> GetAsync(HttpContext context)
             var metadata = new ImageMetadata(fileInfo.LastModified.UtcDateTime, fileInfo.Length);
             return Task.FromResult<IImageResolver>(new PhysicalFileSystemResolver(fileInfo, metadata));
```
ronaldbarendse, line 85: I would recommend renaming `PhysicalFileSystemResolver` into `FileInfoImageResolver`, as it's resolving the image from an `Microsoft.Extensions.FileProviders.IFileInfo` and that's not specific to the `PhysicalFileProvider`.

ronaldbarendse (changes requested):
These changes look good, although it's probably worth noting that if you're using a `ProviderRootPath` that's outside of your webroot, you'll also need to [configure the static file middleware to be able to serve the original file](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/static-files?view=aspnetcore-6.0#serve-files-outside-of-web-root) (requested without commands). Otherwise, only images processed by ImageSharp.Web are returned 😞 

And I'd argue that this would be a breaking change, as users could depend on ImageSharp.Web being able to transparently load/read files that are exposed by the `WebRootProvider`, like Umbraco 9.3.0 does: https://github.com/umbraco/Umbraco-CMS/pull/11783.

I would be fine with this implementation being the default, but then also provide a `WebRootImageProvider` for backwards compatibility. This implementation could then re-use the existing `PhysicalFileSystemResolver` (that IMHO should be renamed to `FileInfoImageResolver`, see comment).

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: The type is explicitly using the `PhysicalFileProvider`

https://github.com/SixLabors/ImageSharp.Web/pull/214/files#diff-d17afca30cd91fe17418d73800c1e3003d61ac7364355e1fd1809acbdcc8def3R57-R59

JimBobSquarePants: I'm not going to introduce a `WebRootProvider`. The current `PhysicalFileSystemProvider` is backwards compatible with out of the box behavior. Umbraco requires significant startup configuration to work as required already and writing providers is trivial enough..

ronaldbarendse: It's not only Umbraco that's using the `WebRootFileProvider` to provide access to static assets though! This change also breaks all [static assets referenced from a Razor Class Library](https://docs.microsoft.com/en-us/aspnet/core/razor-pages/ui-class?view=aspnetcore-6.0&tabs=visual-studio#consume-content-from-a-referenced-rcl-1) (during development/when not running the published application).

The default ASP.NET Core `WebHost` [calls `UseStaticWebAssets()` on a development environment](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/DefaultBuilder/src/WebHost.cs#L212-L220), which uses the same [`CompositeFileProvider` technique](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Hosting/Hosting/src/StaticWebAssets/StaticWebAssetsLoader.cs#L35-L43) to ensure the RCL static assets are served from `_content/{PACKAGE_ID}/`, without having to copy files into your project (which clutters the project, might get updated, deleted, etc.). When running `dotnet publish`, an MSBuild target copies the RCL static assets to your output directory, so it also works on non-development environments (and without requiring the `CompositeFileProvider`).

**To be very explicit about this: RCLs are supported on ASP.NET Core 6 out of the box and this PR _does_ break existing behavior!**

I've updated the sample project to .NET 6, added both an embedded file and RCL project reference on the [`feature/rcl-support`](https://github.com/ronaldbarendse/ImageSharp.Web/tree/feature/rcl-support) branch, so you can test this yourself. And you can also see it working again by uncommenting the [`.AddProvider<WebRootImageProvider>`](https://github.com/ronaldbarendse/ImageSharp.Web/blob/c0669a70f4929f875aee32a172252df66413b62b/samples/ImageSharp.Web.Sample/Startup.cs#L44) (which uses the existing [`PhysicalFileSystemResolver` by the way](https://github.com/ronaldbarendse/ImageSharp.Web/blob/c0669a70f4929f875aee32a172252df66413b62b/src/ImageSharp.Web/Providers/WebRootImageProvider.cs#L69)).

I just can't wrap my head around why you did include a `LegacyV1CacheKey` implementation purely for backwards compatibility with V1 users, but don't want to include a `WebRootImageProvider` for the exact same reason (to allow switch back to the previous behavior). I know writing this provider is trivial, but why not include it by default then? And by doing so, highlighting the weird naming of the `PhysicalFileSystemResolver` again 😉

JimBobSquarePants: It doesn’t break behaviour in a manner that cannot be easily fixed via configuration. If you are storing images outside of the web root you have to configure the static handler. Which most people simply won’t be doing. The provider is command only anyway. 

I HAD to include the legacy cache key as ALL of the replacements break all caches. That will mean silently breaking millions of cached images for some. That’s a profound difference. 

As to why I don’t add one myself. Why SHOULD I? What incentive do I have to add and maintain MORE code with NO BENEFIT to myself.

You need to understand. The maintenance burden for these libraries is MASSIVE. There’s very few bodies of work out there as complicated as the Six Labors libraries maintained by so few. That’s why this middleware was specifically designed to be easily extensible. I give you (figuratively) the tools to easily build items for your requirements and your (figuratively) own paid time.

JimBobSquarePants: The resolver is not oddly named. It is the resolver for the physical provider. They come in pairs, I explained this already.

ronaldbarendse: > As to why I don’t add one myself. Why SHOULD I? What incentive do I have to add and maintain MORE code with NO BENEFIT to myself.
> 
> You need to understand. The maintenance burden for these libraries is MASSIVE. There’s very few bodies of work out there as complicated as the Six Labors libraries maintained by so few. That’s why this middleware was specifically designed to be easily extensible. I give you (figuratively) the tools to easily build items for your requirements and your (figuratively) own paid time.

I totally understand your view on this, but you're actually introducing more code to maintain and making it harder to configure correctly. ASP.NET Core already configures a `PhysicalFileProvider` based on the `WebRootPath`, but only when that folder exists: https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Hosting/Hosting/src/Internal/HostingEnvironmentExtensions.cs#L31-L58.

So I have the following major issues with the current implementation:
1. If the `wwwroot` directory doesn't exist, the `WebRootPath` is not set (and the `WebRootFileProvider` is set to a `NullFileProvider` instance) and this would currently throw exceptions in both `PhysicalFileSystemProvider` and `PhysicalFileSystemCache`;
2. If the exceptions would be fixed, the current `PhysicalFileSystemProvider` creates the missing directory, which would result in a different `WebRootFileProvider` to be assigned on the next startup. There's also no reason for requiring write-access to create the directory, as this provider is read-only (it makes sense for `PhysicalFileSystemCache` to create the cache directory though, as that _does_ require write-access);
3. If you'd fix both these issues, you'll end up with almost the same code as in ASP.NET Core (an thus have introduced additional code to maintain);
4. And because this provider command only, end-users would indeed need to configure additional static file serving to also make the original files available from a directory outside of the web root, which makes it harder to setup correctly. This would again increase the maintenance support, as it might require additional support and/or documentation.

I hope you also understand that I'm trying to help improve the OOTB experience of this library, so everyone can benefit from the time I've put into this, instead of having to (re-)implement core functionality outside of this library myself. No hard feelings though, as you're the maintainer and have the final say, but just want to express my intentions/views on this as well!

JimBobSquarePants: 1. This is by design. If a folder is not explicitly set and the default is not set then it should through. You need to stop coupling the  provider/cache mentally with the ASP.NET Core providers. They're separate and distinct and work well out of the box. (Removing wwwroot is not a common activity)
2. This is a scenario that can never exist because we do and should throw.
3. They're not issues. See comment above about mental coupling.
4. See comment above about mental coupling.

What I think we should do actually. Is strip out all references to `PhysicalFileProvider ` in both the provider and cache and simply use `File`. That way they are obviously distinct and we can move on.

PR #216, opened 2022-02-03
Build against .NET 5 and .NET 6

ronaldbarendse:
### Prerequisites

### Description
I noticed ImageSharp already added .NET 5 and .NET 6 builds in https://github.com/SixLabors/ImageSharp/pull/1745, but ImageSharp.Web is still only built against .NET Core 2.1 (support ended on August 21, 2021) and .NET Core 3.1 (support ends December 3, 2022, see [MS support policy](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)).


ronaldbarendse:
FYI GitHub Actions soon won't be able to build .NET Core 2.1 projects anymore: https://github.com/actions/virtual-environments/issues/4871. Because it's already out of support, I recommend dropping this framework version before v2 ships.

JimBobSquarePants:
I won’t be dropping it while it’s still supported. V2 is designed as a baseline so I can drop targets with V3. (Same principle upstream with ImageSharp. 

Regarding this PR. You’ll need to wire things up for testing against .NET 5/6 in the same manner as ImageSharp

Nuklon:
> I won’t be dropping it while it’s still supported. V2 is designed as a baseline so I can drop targets with V3. (Same principle upstream with ImageSharp.

2.1 is already EOL and unsupported since August 21, 2021, and in 11 days it's also gone from GH Actions.

ronaldbarendse:
> 2.1 is already EOL and unsupported since August 21, 2021, and in 11 days it's also gone from GH Actions.

Unsupported by Microsoft yes, but that doesn't stop people from still using it 😉 .NET Core 2.1 won't be included on the default OS images of GitHub Actions, but that isn't a problem as this build script already installs all required .NET SDKs...

JimBobSquarePants (approved):
This is great. Thanks!

PR #217, opened 2022-02-04
Respect EXIF orientation when resizing

JimBobSquarePants:
### Prerequisites

### Description

In addition to #215 we need the ability to resize images containing EXIF orientation metadata that specifies rotation based upon the users [observed expectation](https://twitter.com/James_M_South/status/1489219049532067842) without updating the EXIF metadata.

This PR updates the `ResizeWebProcessor` to swap dimensions parsed from command when a value indicating EXIF rotation is detected. This behavior can be turned off per request by passing the command `orient=false` (_Note I've deliberately not used an `r` prefix to make this more reusable for other potential transforming processors_).

cc\ @ronaldbarendse

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,160 +1,172 @@
 name: Build
 
 on:
-    push:
-        branches:
-            - master
-        tags:
-            - "v*"
-    pull_request:
-        branches:
-            - master
+  push:
+    branches:
+      - master
+    tags:
+      - "v*"
+  pull_request:
+    branches:
+      - master
 
 jobs:
-    Build:
-        strategy:
-            matrix:
-                options:
-                    - os: macos-latest
-                      framework: netcoreapp3.1
-                      runtime: -x64
-                      codecov: false
-                    - os: ubuntu-latest
-                      framework: netcoreapp3.1
-                      runtime: -x64
-                      codecov: false
-                    - os: windows-latest
-                      framework: netcoreapp3.1
-                      runtime: -x64
-                      codecov: true
-                    - os: windows-latest
-                      framework: netcoreapp2.1
-                      runtime: -x64
-                      codecov: false
+  Build:
+    strategy:
+      matrix:
+        options:
+          - os: macos-latest
+            framework: netcoreapp3.1
+            runtime: -x64
+            codecov: false
+          - os: ubuntu-latest
+            framework: netcoreapp3.1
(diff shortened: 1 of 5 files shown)


JimBobSquarePants:
Question. Do we even need #215 anymore? I'm thinking we actually don't. It's probably best we don't provide a destructive operation - People can implement their own should they need to.

ronaldbarendse:
> Note I've deliberately not used an `r` prefix to make this more reusable for other potential transforming processors

Totally agree and to make it more reusable, the logic should probably also be in a separate helper class. Maybe it should even be a part of ImageSharp itself, so you can easily get the auto-oriented size (and/or crop coordinates) when doing manual transforms?

I see you've already started on related changes in https://github.com/SixLabors/ImageSharp/pull/1976, so maybe add an `OrientationHelper` that accepts sizes, coordinates, rectangles, etc. and transforms them based on the EXIF value? Would also be nice to add a new `ExifSize()` extension method to `ImageInfoExtensions` (or add an overload to `Size(bool autoOrient = true)` that auto-orients the dimensions)?

For ImageSharp.Web the logic could then be moved to `OrientationWebHelper`, which contains the command key (as public const) and related parsing logic that can be used by any web processor that want to be orientation aware (and use the same command to opt-out).

> Question. Do we even need https://github.com/SixLabors/ImageSharp.Web/pull/215 anymore? I'm thinking we actually don't. It's probably best we don't provide a destructive operation - People can implement their own should they need to.

This processor still has a different responsibility: auto-orienting the image for browsers/viewers that don't support auto-rotating based on the EXIF value or when you need to disable this functionality (using `image-orientation: none;` in your CSS), but want the image to be displayed correctly.

JimBobSquarePants:
@ronaldbarendse I don't think moving the current logic into some sort of public helper is worth the maintenance effort to try and design something genuinely useful.  If you want to make your web processors do similar ImageSharp exposes all you need to allow that. 

I've no plans to add such helpers to ImageSharp. This work is an abstraction which operates in the opposite way to the correct methods in ImageSharp.

ronaldbarendse:
Shouldn't the `ranchor` and `rxy` values also respect the EXIF orientation? Otherwise they will select the 'wrong' portion when cropping.

Might be easier to mutate the `ResizeOptions` once, so all parsing methods can be kept clean.

JimBobSquarePants:
@ronaldbarendse Good catch, I'll open a PR today.

PR #220, opened 2022-02-06
Removed the AutoOrientWebProcessor.

dlemstra:
### Prerequisites

### Description
This pull request remove the `AutoOrientWebProcessor` since I suspect that most poeple will use this together with resizing. This is new and has not yet been released so we can still remove it at this point. And incase someone really needs this they can copy the code from this PR.


JimBobSquarePants:
Yep. I considered doing this and held back. I don't think we need it in the repo.

JimBobSquarePants (approved):

PR #221, opened 2022-02-07
Apply updates to XY and Anchor positions based on EXIF metadata

JimBobSquarePants:
### Prerequisites

### Description

Following on from #217  This PR updates `ResizeWebProcessor` to correctly handle `XY` and `Anchor` commands when processing EXIF orientated images.

There's a lot of work involved in handling this with most of the logic based upon some of the work upstream in ImageSharp when normalizing orientation values. The following [image](https://www.daveperrett.com/articles/2012/07/28/exif-orientation-handling-is-a-ghetto/) will help explain the affect on an image each rotation value has.

image

cc @ronaldbarendse

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,4 +1,4 @@
-# How to contribute to ImageSharp.Web
+# How to contribute to SixLabors.ImageSharp.Web
 
 #### **Did you find a bug?**
 
@@ -16,20 +16,18 @@
 
 #### **Do you intend to add a new feature or change an existing one?**
 
-* Suggest your change in the [ImageSharp Gitter Chat Room](https://gitter.im/ImageSharp/General) and start writing code.
+* Suggest your change in the [Ideas Discussions Channel](https://github.com/SixLabors/ImageSharp.Web/discussions?discussions_q=category%3AIdeas) and start writing code.
 
 * Do not open an issue on GitHub until you have collected positive feedback about the change. GitHub issues are primarily intended for bug reports and fixes.
 
 #### **Do you have questions about consuming the library or the source code?**
 
-* Ask any question about how to use ImageSharp in the [ImageSharp Gitter Chat Room](https://gitter.im/ImageSharp/General).
+* Ask any question about how to use SixLabors.ImageSharp.Web in the [Help Discussions Channel](https://github.com/SixLabors/ImageSharp.Web/discussions?discussions_q=category%3AHelp).
 
-#### Code of Conduct
+#### Code of Conduct  
 This project has adopted the code of conduct defined by the [Contributor Covenant](https://contributor-covenant.org/) to clarify expected behavior in our community.
 For more information, see the [.NET Foundation Code of Conduct](https://dotnetfoundation.org/code-of-conduct).
 
-And please remember. ImageSharp.Web is the work of a very, very, small number of developers who struggle balancing time to contribute to the project with family time and work commitments. We encourage you to pitch in and help make our vision of simple accessible imageprocessing available to all. Open Source can only exist with your help.
+And please remember. SixLabors.ImageSharp.Web is the work of a very, very, small number of developers who struggle balancing time to contribute to the project with family time and work commitments. We encourage you to pitch in and help make our vision of simple accessible image processing available to all. Open Source can only exist with your help.
 
 Thanks for reading!
-
-James Jackson-South :heart:
--- a/src/ImageSharp.Web/FormattedImage.cs
+++ b/src/ImageSharp.Web/FormattedImage.cs
@@ -6,6 +6,7 @@ using System.IO;
 using System.Runtime.CompilerServices;
 using SixLabors.ImageSharp.Advanced;
 using SixLabors.ImageSharp.Formats;
+using SixLabors.ImageSharp.Metadata.Profiles.Exif;
 using SixLabors.ImageSharp.PixelFormats;
 
 namespace SixLabors.ImageSharp.Web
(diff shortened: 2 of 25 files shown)


.github/CONTRIBUTING.md
```diff
@@ -1,4 +1,4 @@
-# How to contribute to ImageSharp.Web
```
JimBobSquarePants, line 1: @dlemstra Thought I'd just do this here.

src/ImageSharp.Web/FormattedImage.cs
```diff
@@ -97,6 +98,45 @@ public static FormattedImage Load(Configuration configuration, Stream source)
         /// <param name="destination">The destination stream.</param>
         public void Save(Stream destination) => this.Image.Save(destination, this.encoder);
 
+        /// <summary>
+        /// Returns a value indicating whether the source image contains EXIF metadata for <see cref="ExifTag.Orientation"/>
+        /// indicating that the image is rotated (not flipped).
+        /// </summary>
+        /// <param name="orientation">The decoded orientation. Use <see cref="ExifOrientationMode"/> for comparison.</param>
+        /// <returns>The <see cref="bool"/> indicating whether the image contains EXIF metadata indicating that the image is rotated.</returns>
+        public bool IsExifRotated(out ushort orientation)
```
JimBobSquarePants, line 107: I've made this public as it's useful for other processor implementations. The EXIF parsing logic is identical to `AutoOrientProcessor<T>.GetExifOrientation()`

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
```diff
@@ -182,10 +213,98 @@ or ExifOrientationMode.RightBottom
...
             CultureInfo culture)
-            => parser.ParseValue<AnchorPositionMode>(commands.GetValueOrDefault(Anchor), culture);
+        {
+            AnchorPositionMode anchor = parser.ParseValue<AnchorPositionMode>(commands.GetValueOrDefault(Anchor), culture);
+            if (orientation is >= ExifOrientationMode.Unknown and <= ExifOrientationMode.TopLeft)
+            {
+                return anchor;
+            }
+
+            /*
```
JimBobSquarePants, line 227: I had to use a notepad for this bit. I think my logic is sound but could do with a second opinion.

Jeavon:
I have discovered this PR additionally fixes a null exception for images lacking orientation metadata :thumbsup:

JimBobSquarePants:
Just added a bunch of manual testing to confirm my approach. Seems to work very well.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -32,13 +31,13 @@ public class ImageSharpMiddleware
...
         private static readonly ConcurrentTLruCache<string, ImageMetadata> SourceMetadataLru
-            = new(1024, TimeSpan.FromMinutes(5));
+            = new(1024, TimeSpan.FromSeconds(30));
 
         /// <summary>
         /// Used to temporarily store cache resolver reads to reduce the overhead of cache lookups.
         /// </summary>
         private static readonly ConcurrentTLruCache<string, (IImageCacheResolver, ImageCacheMetadata)> CacheResolverLru
-            = new(1024, TimeSpan.FromMinutes(5));
+            = new(1024, TimeSpan.FromSeconds(30));
```
deanmarcussen, line 40: are you going to pop these back up? or make them configurable?

JimBobSquarePants: I guess we could make them configurable but I've no idea what to name the properties (_InMemoryCacheDuration_?)

I found that if I deleted the cache folder when something was stored in there it prompted a null reference exception in the cache resolver. I dropped the times to make that less likely.

JimBobSquarePants: Ah... They're private static. We can't pass options.

deanmarcussen: > Ah... They're private static. We can't pass options.

True, there might be an argument for making them non static, for performance.
In our scenario, we have multiple instances of the middleware (per tenant), so multiple instances of the cache per tenant, is perhaps desirable, as the items count will be less, so the lookup a little faster.
(then becomes configurable)

Something we can think about later though :)

src/ImageSharp.Web/FormattedImage.cs
ronaldbarendse: Why is this only checking for rotated images and not flipped ones? I think that made sense when only considering width/height, but because the `ResizeWebProcessor` also needs to handle the correct X/Y and anchor position, this should also return `true` when it's flipped.

I'd also rename it to `TryGetExifOrientation(out ushort orientation)` and return whether the image contains a valid EXIF orientation value or not. Then just check if it's false or the orientation is set to 1 (top/left) before doing any of the transforms.

JimBobSquarePants: Why is this only checking for rotated images and not flipped ones? 

Because rotated is an important metric as that affects Size. Flipped does not affect that property and I wanted a quick way to tell, writing the method to reflect that desire. Can easily refactor though to suit your design and move rotate the check inside the processor itself.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
JimBobSquarePants: Yeah that might well be an idea then.

ronaldbarendse (changes requested):
I've added my suggestions in PR https://github.com/SixLabors/ImageSharp.Web/pull/224, as the current implementation doesn't correctly parse/transform the center coordinates and isn't easily re-usable in custom processors.

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
```diff
@@ -172,7 +162,51 @@ or ExifOrientationMode.RightBottom
...
 
-            return new PointF(coordinates[0], coordinates[1]);
+            PointF center = new(coordinates[0], coordinates[1]);
+            if (orientation is >= ExifOrientationMode.Unknown and <= ExifOrientationMode.TopLeft)
+            {
+                return center;
+            }
+
+            AffineTransformBuilder builder = new();
+            Size size = image.Image.Size();
```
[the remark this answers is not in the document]
JimBobSquarePants: I'd actually forgotten that those coordinates were 0-1. They're not documented well on the underlying type. Suggested fixes are much more useful here that in a separate PR.

I don't want the other changes. I don't want to have to create yet another public type, especially one that is mostly filled with methods specific to `ResizeWebProcessor`. The `PointF` Transform method for example requires the X/Y properties to be within a 0-1 range. Not useful for anything that currently exists. You're creating solutions to problems I don't and won't have. Processor implementation outside of this library are the responsibility of whoever develops them not me.  I'm building a framework, not a utility belt.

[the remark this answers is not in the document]
JimBobSquarePants: I've implemented a proper fix for XY transforms now anyway, took me longer than I'm proud of to figure out. If you look at the image above and compare it to the code the process actually becomes pretty clear.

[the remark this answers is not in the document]
ronaldbarendse: Most of those methods are specific to the `ResizeOptions` indeed, which can be re-used on different web processors that require resizing (e.g. when trying to crop to a specific target rectangle, which Umbraco does in the `CropWebProcessor` based on crop coordinates), so it's not really specific to `ResizeWebProcessor`.

Making the `PointF` transform use a different range would be as simple as moving the `Size` variable to a parameter, but I get the point you're making here 😉

I've quickly tested the latest changes and it all seems to work as expected 👍🏻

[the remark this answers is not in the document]
JimBobSquarePants: Thanks for testing, that's very useful 👍

PR #222, opened 2022-02-07
Load source images asynchronously to prevent thread pool exhaustion

kroymann:
### Prerequisites

### Description
While debugging a memory dump taken during an outage of our production image processing servers, we discovered that nearly all of the thread pool worker threads were blocked waiting on network IO loading source image data, which led to thread pool starvation that was preventing any async work from executing.  This PR addresses the source of the problem by replacing the synchronous `Image.Load()` call with the equivalent async call.

### DIFF
--- a/src/ImageSharp.Web/FormattedImage.cs
+++ b/src/ImageSharp.Web/FormattedImage.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
 using SixLabors.ImageSharp.Advanced;
 using SixLabors.ImageSharp.Formats;
 using SixLabors.ImageSharp.PixelFormats;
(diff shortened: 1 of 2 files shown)


sebastienros:
Should the non-async ones now be made obsolete such that next version (assuming it will be a major one?) can remove them?

kroymann:
> Should the non-async ones now be made obsolete in the next version (assuming it will be a major one?)

Probably yes?  I'd be willing to be that the vast majority of the time, the image source stream requires network or file I/O to read, and therefore giving people the ability to synchronously load it is almost always going to be the wrong thing to do.

deanmarcussen:
Looks like you could use SaveAsync as well?

kroymann:
> Looks like you could use SaveAsync as well?

I originally intended to do that, but then realized that the `outStream` is a `RecyclableMemoryStream` which means both the source and destination are in-memory buffers and therefore the save operation shouldn't need to do any asynchronous IO.

JimBobSquarePants (approved):
Perfect, thanks! 

We’ll leave the old methods in for now without obsoleting them. I doubt anyone uses them but I think I’d prefer to maintain the sync/async API balance since your using `Save()`

PR #224, opened 2022-02-08
Move re-usable orientation logic into OrientationHelper

ronaldbarendse:
### Prerequisites

### Description
This PR moves the orientation related parsing/transforms into a re-usable `OrientationHelper` and also fixes the transforms done on the center coordinates.


JimBobSquarePants:
You really should have asked first chum. I don’t want to do this. All the size/anchor/position logic belongs in the resize processor and is not reusable

PR #225, opened 2022-02-08
Replace PhysicalFileProvider with File in PhysicalFileSystemCache and PhysicalFileSystemProvider

ronaldbarendse:
### Prerequisites

### Description
As per request on https://github.com/SixLabors/ImageSharp.Web/pull/214#discussion_r801533015, this strips out all references to `PhysicalFileProvider` and replaces them with `File` 👍🏻

The `PhysicalFileSystemProvider` might need some additional path normalization, although I think ASP.NET Core already normalizes the `context.Request.Path`: this is something to double check...


src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -72,16 +65,16 @@ public bool IsValidRequest(HttpContext context)
         /// <inheritdoc/>
         public Task<IImageResolver> GetAsync(HttpContext context)
         {
-            // Path has already been correctly parsed before here.
-            IFileInfo fileInfo = this.fileProvider.GetFileInfo(context.Request.Path.Value);
+            string path = Path.Join(this.providerRootPath, context.Request.Path.Value);
```
JimBobSquarePants, line 68: I _think_ you're correct here and the path is already normalized since it's specialized `PathString`. [PhysicalFileProvider](https://source.dot.net/#Microsoft.Extensions.FileProviders.Physical/PhysicalFileProvider.cs,228) does a bunch of extra checks because it does not have that guarantee.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -44,18 +43,12 @@ public class PhysicalFileSystemProvider : IImageProvider
             FormatUtilities formatUtilities)
         {
             Guard.NotNull(environment, nameof(environment));
-
-            // ContentRootPath is never null.
-            // https://github.com/dotnet/aspnetcore/blob/b89eba6c3cda331ee98063e3c4a04267ec540315/src/Hosting/Hosting/src/WebHostBuilder.cs#L262
```
JimBobSquarePants, line 49: This comment exists to indicate why we do not do a null check for `environment.ContentRootPath`.

src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
```diff
@@ -33,7 +32,7 @@ public PhysicalFileSystemCacheResolver(IFileInfo metaFileInfo, FormatUtilities f
         /// <inheritdoc/>
         public async Task<ImageCacheMetadata> GetMetaDataAsync()
         {
-            using Stream stream = this.metaFileInfo.CreateReadStream();
+            using Stream stream = this.metaFileInfo.OpenRead();
```
JimBobSquarePants, line 35: `OpenRead` will allocate a buffer. We need to copy the implementation [here](https://source.dot.net/#Microsoft.Extensions.FileProviders.Physical/PhysicalFileInfo.cs,46).

src/ImageSharp.Web/Resolvers/PhysicalFileSystemResolver.cs
```diff
@@ -30,6 +29,6 @@ public PhysicalFileSystemResolver(IFileInfo fileInfo, in ImageMetadata metadata)
         public Task<ImageMetadata> GetMetaDataAsync() => Task.FromResult(this.metadata);
 
         /// <inheritdoc/>
-        public Task<Stream> OpenReadAsync() => Task.FromResult(this.fileInfo.CreateReadStream());
+        public Task<Stream> OpenReadAsync() => Task.FromResult<Stream>(this.fileInfo.OpenRead());
```
JimBobSquarePants, line 32: Same change required as the cache resolver.

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
```diff
@@ -54,18 +48,13 @@ public class PhysicalFileSystemCache : IImageCache
             FormatUtilities formatUtilities)
         {
             Guard.NotNull(environment, nameof(environment));
-            Guard.NotNull(options, nameof(options));
```
JimBobSquarePants, line 57: We still need the null check for options.

JimBobSquarePants (changes requested):
Thanks for doing this.

Looking good, only some performance tweaks and sanitation stuff to fix  I reckon.

JimBobSquarePants:
I wonder what is happening with the tests? Do they pass locally?

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
ronaldbarendse: I've updated this code to ensure it also works when `environment.WebRootPath` is null (and in that case use the same `wwwroot` default as [ASP.NET Core uses](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Hosting/Hosting/src/Internal/HostingEnvironmentExtensions.cs#L31-L44)). This removes the need for any null checks here...

src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
ronaldbarendse: Done in https://github.com/SixLabors/ImageSharp.Web/pull/225/commits/bc0ac62da85458aa3c8d19b0f3b9ff48e7c01359 👍🏻

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
ronaldbarendse: I removed this check, because the code below did a conditional null check:

https://github.com/SixLabors/ImageSharp.Web/blob/f97516e3a3461c2e3c0a3d10236de70e1bcaa63b/src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs#L53-L54

The options will always be registered when using `builder.AddImageSharp()` though, so I've added the guard again and removed this custom logic...

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
ronaldbarendse: I've added some path normalization, just to be sure we're not allowing access below the configured provider root path 👍🏻

ronaldbarendse:
> I wonder what is happening with the tests? Do they pass locally?

They did pass locally yes, so might just be a flaky test run 🤔 I've incorporated all requested changes, so let's see if the tests now complete without any issues 🤞🏻

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
JimBobSquarePants: Ah.... Actually I'm not sure about that. I can't see where the options would automatically be created.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -94,13 +87,23 @@ public Task<IImageResolver> GetAsync(HttpContext context)
         /// <returns><see cref="string"/> representing the fully qualified provider root path.</returns>
         internal static string GetProviderRoot(PhysicalFileSystemProviderOptions providerOptions, string webRootPath, string contentRootPath)
         {
-            string providerRoot = string.IsNullOrWhiteSpace(providerOptions.ProviderRootPath)
-                ? webRootPath
-                : providerOptions.ProviderRootPath;
-
-            return Path.IsPathFullyQualified(providerRoot)
+            string providerRoot = providerOptions.ProviderRootPath ?? webRootPath ?? "wwwroot";
+            string fullPath = Path.IsPathFullyQualified(providerRoot)
```
JimBobSquarePants, line 91: Here's where is gets wacky. 
https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Hosting/Hosting/src/Internal/HostingEnvironmentExtensions.cs#L31-L63

If `PhysicalFileSystemProviderOptions.ProviderRootPath` and `WebHostOptions.WebRoot` are both null and `wwwroot` doesn't exist then setting `wwwroot` as a fallback will mean that our provider then points to a folder that doesn't exist and all images requests will silently fail.

I would suggest we throw here if both values are null.

ronaldbarendse: In that case the original image can't be found/doesn't exist and nothing will happen... I'd rather see it update the `Match` callback to `false`, so it doesn't even try to use this provider anymore!

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -43,19 +42,10 @@ public class PhysicalFileSystemProvider : IImageProvider
...
-            // Allow configuration of the provider without having to register everything
-            PhysicalFileSystemProviderOptions providerOptions = options != null ? options.Value : new();
-            string providerRootPath = GetProviderRoot(providerOptions, environment.WebRootPath, environment.ContentRootPath);
-
-            // Ensure provider directory is created before initializing the file provider
-            Directory.CreateDirectory(providerRootPath);
-            this.fileProvider = new PhysicalFileProvider(providerRootPath);
+            this.providerRootPath = GetProviderRoot(options.Value, environment.WebRootPath, environment.ContentRootPath);
             this.formatUtilities = formatUtilities;
         }
```
ronaldbarendse, line 50: As mentioned in https://github.com/SixLabors/ImageSharp.Web/pull/225#discussion_r802467562, this would disable the provider if the provider root directory doesn't exist:

```suggestion

            // Disable provider if root path doesn't exist
            if (!Directory.Exists(this.providerRootPath))
            {
                this.Match = _ => false;
            }
        }
```

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
ronaldbarendse: Looks like registration isn't needed anyway: https://stackoverflow.com/questions/55074806/why-ioptions-is-getting-resolved-even-if-not-registered

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: That still means it silently fails. You want to capture configuration errors at startup with explicit messages. This could be the only provider installed just misconfigured.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: No, that’s not wise as described above. If my provider isn’t matching with no explanation that’s an afternoon of debugging wasted

src/ImageSharp.Web/Caching/PhysicalFileSystemCache.cs
JimBobSquarePants: TIL !!

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
ronaldbarendse: OK, so it should be a little more nuanced: when a specific root provider path is configured that doesn't exist, it should try to create that directory (like [ASP.NET also does](https://github.com/dotnet/aspnetcore/blob/c85baf8db0c72ae8e68643029d514b2e737c9fae/src/Hosting/Hosting/src/Internal/HostingEnvironmentExtensions.cs#L51)), which would then throw an exception if it's invalid...

Having the default `IImageProvider` throwing exceptions when you've not done any customizations to ImageSharps configuration would be a bad experience in my opinion.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: There's profound difference here. 

We're not the framework which is explicitly designed to work without the static folder as that folder and static files are not a requirement. 

We're the polar opposite. If someone want to add ImageSharp.Web they have to explicitly register the library at startup and that expresses intent. They intend to serve images via the middleware and have registered the physical provider (implicitly or explicitly). That provider needs a location from which to serve the images in order to function. If they have not provided a location and have deleted the wwwroot folder then that becomes a case of misconfiguration because they have expressed intent by deleting the default folder.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: You've missed my point I think. Hopefully my additional explanation above makes sense.

I do agree that the provider should try to create the directory if it is explicitly set and doesn't exist.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
ronaldbarendse: Fair points, I've updated (and simplified) both the provider and cache implementations to throw an exception when the directory can't be determined (either not explicitly set in the options or the `WebRootPath` is not set, which would indicate the `wwwroot` directory doesn't exist).

ronaldbarendse: After refactoring this to the current state, I actually think it indeed makes sense that users either have to explicitly configure the provider/cache root path _or_ ensure the web root path is set (which will be the case when the default wwwroot directory exists or it's also explicitly set).

The only thing that might need to be reconsidered is whether ImageSharp should ensure those root directories exist/are created inside the constructors (doing expensive operations and causing side-effects inside constructors is considered bad practice). This also shouldn't be required, as the provider won't find any files inside the empty directory and the cache will create it before writing files anyway...

JimBobSquarePants:
@ronaldbarendse I haven't forgotten this, just needed a bit of a break.

JimBobSquarePants:
I still think that provider should be implemented on demand by developers themselves. History has proven it will be misused due to a lack of understanding of how file system providers work and I’ll get poor issues raised.

Alerinos:
> I still think that provider should be implemented on demand by developers themselves. History has proven it will be misused due to a lack of understanding of how file system providers work and I’ll get poor issues raised.

I propose to create the possibility of setting up several suppliers. Quite often we have a case that we use a different supplier for different purposes. wwwroot for static files, public for files uploaded by the user, privateRoot for e.g. private files available in the administration panel. Let the programmer determine how many suppliers there are and where they should be.

deanmarcussen:
> I propose to create the possibility of setting up several suppliers. Quite often we have a case that we use a different supplier for different purposes. wwwroot for static files, public for files uploaded by the user, privateRoot for e.g. private files available in the administration panel. Let the programmer determine how many suppliers there are and where they should be.

I agree. This is exactly what you should do. The `IImageProvider` is a very simple interface, so for custom requirements like you have, you should implement one that suits you, and maps to paths you find approriate for your configuration.

JimBobSquarePants:
@Alerinos As @deanmarcussen describes this is pre-existing behavior. You can [implement and register](https://docs.sixlabors.com/articles/imagesharp.web/imageproviders.html?tabs=tabid-1%2Ctabid-1a) an unlimited count of custom providers.

>Quite often we have a case that we use a different supplier for different purposes.

Who is "we" in this case and do they have [Six Labors](https://sixlabors.com/pricing/) licenses? You've already been asked to [not ask support question on PRs](https://github.com/SixLabors/ImageSharp.Web/pull/225#issuecomment-1062383372) and it strikes me as very odd that you have immediately repeated the behavior.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: I would tend to agree but wouldn't say these are particularly expensive operations. They're basically null checks and some string mangling.

JimBobSquarePants (approved):
I think this is good to go in. Thanks for your assistance! 👍

PR #235, opened 2022-03-08
Allow ImageSharp to determine the best pixel format on load.

JimBobSquarePants:
### Prerequisites

### Description

This one's for @antonfirsov 

This is a breaking change that massively optimizes memory usage for most common scenarios by allowing individual `IImageWebProcessor` instances to return a value indicating whether an alpha component is required in order to correctly process the image based upon the commands passed to that processor. 

If an `IImageWebProcessor.RequiresAlphaAwarePixelFormat(CommandCollection, CommandParser, CultureInfo)` instance call returns `true` then we load an `Image<Rgba32>` as previously. If however, it returns `false` then we simply load `Image`. This allows individual image formats to deliver the best pixel format based upon the encoded input stream.

Control of the returned value can be fine-grained. For example, `ResizeWebProcessor` will only return `true` if it receives commands indicating that a `BoxPad` or `Pad` resize operation should be performed since they can have the background color set.

I've also thrown in some cleanup here to fix up the unit tests as the way they were constructed was causing issues against he non-threadsafe AWS test shim.


JimBobSquarePants:
I'll add some tests to bring coverage up tomorrow

tocsoft:
Is the alpha channel enough? what about grayscale source images? there is currently no way that I can see for processor to tell us they plan on doing things like drawing a 2nd image frame over the top etc.

JimBobSquarePants:
Yeah should be enough. The idea is that individual processors will instruct the middleware whether an alpha channel is required so a custom drawing processor would always return true unless the developer wanted more fine grained control.

JimBobSquarePants:
OK. I have this working and am happy with the result. Ignore the code coverage issues around LRU cache. I'm gonna open another PR to deal with that.

tocsoft (approved):
not a huge fan of the name `RequiresAlphaAwarePixelFormat()` as it more than just alpha it also impacts other things too, user uploading a PNG in L8 and a processor wanting water mark the image with red text they would also need to enable this flag otherwise it'll end up grayscale but none of that output is anything to do with Alpha as it'd work just as well on Rgb24 as well as Rgba32.

JimBobSquarePants:
Agreed there.... What would you call it then? I've tried two names so far and one was only marginally better than the other 😵😅

"RequiresDeepColorPixelFormat" perhaps?

https://en.wikipedia.org/wiki/Color_depth#Deep_color_(30-bit)

Or should we return a bit depth enum?

```c#
Default
Bit24
Bit32
```

>user uploading a PNG in L8 and a processor wanting water mark the image with red text they would also need to enable this flag otherwise it'll end up grayscale

It's actually impossible with the current architecture to change the color format of your input image unless you adjusted the metadata in one of the event handlers. e.g. `OnProcessedAsync` so if an input was gray and someone wanted to add a colored watermark then they would have to do that.

tocsoft:
I don't think we actually need to worry about 24 bit depths I was using that to demonstrate the difference really. 

Some options off the top of my head:
- `RequiresWideColorPixelFormat`
- `RequiresFullColorPixelFormat`
- `RequiresExpandedColorPixelFormat`

or inverted logic and call it 
- `SupportsNaturalPixelFormat`
- `SupportsAnyPixelFormat` 
- `SupportsMeoryOptimisedImages`

JimBobSquarePants:
After a bit of research I ended up not inverting the logic and simply renaming the method. According to https://en.wikipedia.org/wiki/Color_depth#True_color_(24-bit) `RequiresTrueColorPixelFormat` is accurate for 32bit RGBA. 

`SupportsNaturalPixelFormat` would have been my favorite choice for the inverted approach but I felt there might be some sense of ambiguity from developers implementing the interface as to what that meant in technical terms.

PR #240, opened 2022-03-22
Allow runtime physical deletion of cached images

### LINKED ISSUE #239: Fix exception when cached file is deleted, but still in the LRU cache
We should fix whatever issue is causing the LRU cache to throw if a cache item is returned that is then null. It's not supposed to do that.

Edit. A quick look at the code suggests that the error is not thrown in the cache but later in `SendResponseAsync`. I could be wrong though. I'll do a thorough investigation and add a fix now.

_Originally posted by @JimBobSquarePants in https://github.com/SixLabors/ImageSharp.Web/discussions/237#discussioncomment-2414118_

JimBobSquarePants:
### Prerequisites

### Description

Allows the runtime deletion of cached images by handling any exceptions thrown during resolving the cached image. Fix #239 

If an exception is thrown we do a complete run through of the processing again to ensure that any changes to both the source and image are captured. If an issue still occurs we'll log and rethrow.

CC @ronaldbarendse

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageContext.cs
+++ b/src/ImageSharp.Web/Middleware/ImageContext.cs
@@ -86,9 +86,12 @@ namespace SixLabors.ImageSharp.Web.Middleware
         }
 
         /// <summary>
-        /// Returns the current HTTP request display url.
+        /// Returns the current HTTP image request display url.
         /// </summary>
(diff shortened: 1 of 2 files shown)


ronaldbarendse (approved):
Looks good to me!

PR #241, opened 2022-03-25
Expose ExifOrientationUtilities and ICommandConverter implementations

JimBobSquarePants:
### Prerequisites

### Description

This is a rework of #224 which exposes utilities for transforming various types based upon the EXIF orientation value contained
within an image EXIF metadata. In addition I've made the `ICommandConverter<T>` implementations public since they allow users to more easily unit test their own `IImageWebProcessor` implementations.

There's no functional changes other than `ExifOrientationUtilities.Transform(Vector2...)` will normalize the input position based upon a given min/max range so is more flexible. 

I've deliberately kept the parsing of values from commands separate for now as it feels very specific to the `ResizeWebProcessor` and the code can be easily copied if required (_it's not complex_).

cc @ronaldbarendse


JimBobSquarePants:
@deanmarcussen This is likely the last PR before I release V2.0.0 would it be possible for you to do a performance run (or show me docs so I can refresh my memory how to do it myself?)

ronaldbarendse:
Thanks for making this utility class public, as Umbraco will also need to use the same transforms to convert crop coordinates to a crop rectangle. I did notice some strange outputs though, which I can reproduce (on main branch) by changing to `rxy=1,1` in the sample project:

Landscape images

Portrait images

These images should all resize/crop to the bottom right corner, but that's currently not the case...

JimBobSquarePants:
>I did notice some strange outputs though, which I can reproduce (on main branch) by changing to rxy=1,1 in the sample project:

Awww maaaan! I thought I'd cracked this! That's frustrating.

JimBobSquarePants:
@ronaldbarendse That's it fixed now.

ronaldbarendse:
Awesome! I've tested the new logic with the `CropWebProcessor` and can confirm this is now working correctly using the following code:

```c#
// Values from crop command (0-1)
float left, top, right, bottom;

// Transform left/top and right/bottom points
ushort orientation = GetExifOrientation(image, commands, parser, culture);
Vector2 xy1 = ExifOrientationUtilities.Transform(new Vector2(left, top), Vector2.Zero, Vector2.One, orientation);
Vector2 xy2 = ExifOrientationUtilities.Transform(new Vector2(right, bottom), Vector2.Zero, Vector2.One, orientation);

// Get correct points (as they can get swapped) and convert to pixel based rectangle
Size size = image.Image.Size();
var cropRectangle = Rectangle.FromLTRB(
    (int)MathF.Round(MathF.Min(xy1.X, xy2.X) * size.Width),
    (int)MathF.Round(MathF.Min(xy1.Y, xy2.Y) * size.Height),
    (int)MathF.Round(MathF.Max(xy1.X, xy2.X) * size.Width),
    (int)MathF.Round(MathF.Max(xy1.Y, xy2.Y) * size.Height));
```

Maybe transforming and scaling the values could also be done in a single matrix/vector operation, but getting to the above code was already a goal achieved 😉 

Also good to see you're exposing the converters to help unit testing, can you also do the same for the `FormattedImage` constructor? I see that's also used in the ImageSharp.Web tests and Umbraco currently resorts to reflection to do the same for testing the `CropWebProcessor`: https://github.com/umbraco/Umbraco-CMS/blob/ad66d61b0403cb7c8bf760467eb841a160206c8a/tests/Umbraco.Tests.UnitTests/Umbraco.Web.Common/ImageProcessors/CropWebProcessorTests.cs#L47-L78

JimBobSquarePants:
>Maybe transforming and scaling the values could also be done in a single matrix/vector operation, but getting to the above code was already a goal achieved 😉

Ah but you're already scaling the crop rectangle down. If you'd passed the rectangle bounds as `min`, `max` and the `topleft`/`rightbottom `as the actual pixel values you get automatic scaling internally so your results match.

It looks like you could avoid the `round`/`min `operation by using `ExifOrientationUtilities.Transform(Size size, ushort orientation)`. This will always give you the correct size.

I've just made that constructor public for you. 😄

ronaldbarendse:
> Ah but you're already scaling the crop rectangle down. If you'd passed the image bounds as `min`, `max` and the `topleft`/`rightbottom `as the actual pixel values you get automatic scaling internally so your results match.

But then the actual pixel values would be normalized to a 0-1 value again when doing the matrix transform and I'd have to multiply the top-left/right-bottom _and_ set the max value to the right width/height... The `CropWebProcessor` already accepts percentage based coordinates (actually relative distances from all 4 sides), so it's already limited to 0-1, removing the need to know the actual image size.

> It looks like you could avoid the `round`/`min `operation by using `ExifOrientationUtilities.Transform(Size size, ushort orientation)`. This will always give you the correct size. You can also explicitly cast a `RectangleF` to a `Rectangle`. or use `Rectangle.Truncate(RectangleF rectangle)` if you think that looks neater.

I've tried that, but the transform might result in the top-left point not to be the actual top-left point of the rectangle, but instead the top-right, bottom-left or bottom-right one. Using the min/max ensures you get the correct points of the rectangle and rounding instead of truncating seems to give more consistent results.

ronaldbarendse:
Another thing to note: whenever the transform short-circuits on the orientation, the min/max normalisation is skipped. Shouldn't this always be applied, so the position is always between those values?

JimBobSquarePants:
>Another thing to note: whenever the transform short-circuits on the orientation, the min/max normalisation is skipped. Shouldn't this always be applied, so the position is always between those values?

No. If it doesn't require transforming we shouldn't touch the input value. 

I'm gonna have a quick look at he transform function. Think I missed a trick to solve your incorrect output.

JimBobSquarePants:
Just fixed the `Vector2` transform so that it uses the correct dimensions for descaling and added a bunch of tests so you should now be able to do the following.

```c#
// Values from crop command (0-1)
float left, top, right, bottom;

// Transform left/top and right/bottom points
ushort orientation = GetExifOrientation(image, commands, parser, culture);
Vector2 xy1 = ExifOrientationUtilities.Transform(new Vector2(left, top), Vector2.Zero, Vector2.One, orientation);
Vector2 xy2 = ExifOrientationUtilities.Transform(new Vector2(right, bottom), Vector2.Zero, Vector2.One, orientation);

// Scale points to a pixel based rectangle.
Size size = ExifOrientationUtilities.Transform(image.Image.Size());
Rectangle.Round(
    RectangleF.FromLTRB(
        xy1.X * size.Width,
        xy1.Y * size.Height,
        xy2.X * size.Width,
        xy2.Y * size.Height));
```

ronaldbarendse:
> Another thing to note: whenever the transform short-circuits on the orientation, the min/max normalization is skipped. Shouldn't this always be applied, so the position is always between those values?

After reading the tests, I now get why you need a min/max value: not to clamp the input value, but to correctly rotate/flip on a specific plane. It might be helpful to add the following test cases to `TransformVectorData`, as that directly makes this clear:

```c#
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.Unknown, new Vector2(-25F, -25F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.TopLeft, new Vector2(-25F, -25F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.TopRight, new Vector2(175F, -25F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.BottomRight, new Vector2(175F, 125F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.BottomLeft, new Vector2(-25F, 125F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.LeftTop, new Vector2(-25F, -25F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.RightTop, new Vector2(-25F, 175F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.RightBottom, new Vector2(125F, 175F) },
{ new Vector2(-25F, -25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.LeftBottom, new Vector2(125F, -25F) }
```

> so you should now be able to do the following

Unfortunately not, as the `xy1` and `xy2` values still might get transformed to points that are not the top-left and bottom-right points anymore and thus `RectangleF.FromLTRB()` not creating the correct rectangle. Looking at the reference source of [`System.Windows.Rect(Point point1, Point point2)`](https://referencesource.microsoft.com/#windowsbase/Base/System/Windows/Rect.cs,73), this does something similar as well. You might want to add a constructor that accepts 2 points to the `Rectangle` and `RectangleF` structs in ImageSharp to make this easier in the future, but I guess this won't be a very common use-case to support...

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
```diff
@@ -179,46 +177,7 @@ public bool RequiresTrueColorPixelFormat(CommandCollection commands, CommandPars
             }
 
             PointF center = new(coordinates[0], coordinates[1]);
```
ronaldbarendse, line 179: Probably better to directly create a `Vector2` here, instead of implicitly converting from the `PointF`:
```suggestion
            Vector2 center = new(coordinates[0], coordinates[1]);
```

src/ImageSharp.Web/ExifOrientationUtilities.cs
```diff
@@ -0,0 +1,212 @@
...
+        /// <param name="orientation">The EXIF orientation.</param>
+        /// <returns>
+        /// The transformed anchor.
+        /// </returns>
+        public static AnchorPositionMode Transform(AnchorPositionMode anchor, ushort orientation)
+        {
+            if (orientation is >= ExifOrientationMode.Unknown and <= ExifOrientationMode.TopLeft)
+            {
+                return anchor;
+            }
```
ronaldbarendse, line 98: Does this short-circuit really have a performance improvement compared to the switch statements below?

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
```diff
@@ -374,15 +249,5 @@ private static ushort GetExifOrientation(FormattedImage image, CommandCollection
             image.TryGetExifOrientation(out ushort orientation);
             return orientation;
         }
```
ronaldbarendse, line 251: I'd rather use an explicit value when the EXIF orientation can't be retrieved from the image:

```
private static ushort GetExifOrientation(FormattedImage image, CommandCollection commands, CommandParser parser, CultureInfo culture)
{
    if ((commands.Contains(Orient) && !parser.ParseValue<bool>(commands.GetValueOrDefault(Orient), culture)) ||
        !image.TryGetExifOrientation(out ushort orientation))
    {
        return ExifOrientationMode.Unknown;
    }

    return orientation;
}
```

This is (currently) functionally the same, so it's more a styling preference.

JimBobSquarePants:
Those test cases make no sense. The provided vectors are not within the min/max range. Min and max allow proportional scaling of vector dimension lengths. 0-150 becomes 0-1.

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
ronaldbarendse: Another minor styling issue I noticed is that the order of the commands differ between the constants and `ResizeCommands`, plus `Sampler` and `PadColor` can be assigned in the `ResizeOptions` object initializer.

JimBobSquarePants: There’s less codegen in my version

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
JimBobSquarePants: Yeah will update

src/ImageSharp.Web/ExifOrientationUtilities.cs
JimBobSquarePants: Yeah though I should swap the arguments.

ronaldbarendse:
> Those test cases make no sense. The provided vectors are not within the min/max range. Min and max allow proportional scaling of vector dimension lengths. 0-150 becomes 0-1.

I agree that the test cases might not make much sense in the usage context (transforming a point within a plane), but it does show/validate that the min/max values aren't clamping the value, but instead the values 'wrap around' them. I still think that's a valid thing to test...

I'm very happy to see these changes are going to be a part of the v2 release: it's simply awesome to have ImageSharp.Web support EXIF-orientation aware image resizing and provide some utility classes to do the same in your own custom processors! 🎉

tests/ImageSharp.Web.Tests/Helpers/ExifOrientationUtilitiesTests.cs
```diff
@@ -0,0 +1,139 @@
...
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.TopLeft, new Vector2(25F, 25F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.TopRight, new Vector2(125F, 25F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.BottomRight, new Vector2(125F, 75F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.BottomLeft, new Vector2(25F, 75F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.LeftTop, new Vector2(25F, 25F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.RightTop, new Vector2(25F, 125F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.RightBottom, new Vector2(75F, 125F) },
+                { new Vector2(25F, 25F), Vector2.Zero, new Vector2(150, 100), ExifOrientationMode.LeftBottom, new Vector2(75F, 25F) }
```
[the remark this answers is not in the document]
JimBobSquarePants: That's not what this is testing nor should be. I'm testing two things.

 1. That the input values are scaled correctly from `25, 25` to `16.667, .25`. 
 2. That the input values are transformed to match a precalculated expected result. 

See the asci diagram [here](https://github.com/SixLabors/ImageSharp.Web/pull/241/files#diff-04782181c7482ecbfcbbed23d3eec6f4960985bc01828c070db3e24cb993a6acR13-R26). `25, 25` is marked `X` as the tested position

```
        //                  150
        // ┌─────┬─────┬─────┬─────┬─────┬─────┐
        // │     │     │     │     │     │     │
        // │     │    FFFFFFFFFFFFFFF    │     │
        // ├─────X────F┌─────┬─────┬─────┼─────┤
        // │     │    F│     │     │     │     │
        // │     │    FFFFFFFFFFFFFFF    │     │
        // ├─────┼────F┌─────┬─────┬─────┼─────┤ 100
        // │     │    F│     │     │     │     │
        // │     │    F│     │     │     │     │
        // ├─────┼────F├─────┼─────┼─────┼─────┤
        // │     │    F│     │     │     │     │
        // │     │     │     │     │     │     │
        // └─────┴─────┴─────┴─────┴─────┴─────┘
```

Should I be testing a position greater than the centre to ensure my previous issue is covered. Probably but `-25,-25` are non-sensical input values given the scaling bounds of `0,0` and `150,100`. 

TL;DR
The vector position MUST be within the Min/Max vector range.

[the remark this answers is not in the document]
ronaldbarendse: > The vector position MUST be within the Min/Max vector range.

I guess my current issue is that this isn't validated/enforced and the result is unclear. 

Wouldn't it make sense to throw an `ArgumentOutOfRangeException` if the position is outside of the min/max range:
```c#
if (position.X < min.X || position.Y < min.Y)
{
    throw new ArgumentOutOfRangeException(nameof(position), position, $"The position X and Y values should be greater than or equal to the min X and Y values ({min}).");
}

if (position.X > max.X || position.Y > max.Y)
{
    throw new ArgumentOutOfRangeException(nameof(position), position, $"The position X and Y values should be lower than or equal to the max X and Y values ({max}).");
}
```

Or if you don't want any exceptions in this code-path, enforce this by clamping the values by adding this to the top of the method and then use `normalized` instead:
```c#
var normalized = new Vector2(Math.Clamp(position.X, min.X, max.X), Math.Clamp(position.Y, min.Y, max.Y));
```

I'm fine with clamping the input value myself, but just wanted to point this out 😉

------------

From my testing, the following 'out of bounds' center positions currently result in the same 2 crops:
```
/Portrait_0.jpg?width=60&height=50&rxy=0,0
/Portrait_0.jpg?width=60&height=50&rxy=0,-1

/Portrait_0.jpg?width=60&height=50&rxy=0,1
/Portrait_0.jpg?width=60&height=50&rxy=0,2
```

But even through the resulting image is the same, because the URL is different they are cached as 2 separate files. This might not be a big deal, as that is currently also the case when using a negative width or height (and if both are 0 or lower, it will not even resize, but still cache the image), but still something to be aware of. And this can be manually worked-around by removing any invalid commands in the `OnParseCommandsAsync` callback, similar to how the maximum 4000x4000 size is handled...

src/ImageSharp.Web/ExifOrientationUtilities.cs
```diff
@@ -0,0 +1,212 @@
...
+                AnchorPositionMode.BottomLeft => orientation switch
+                {
+                    ExifOrientationMode.TopRight or ExifOrientationMode.RightTop => AnchorPositionMode.BottomRight,
+                    ExifOrientationMode.BottomLeft or ExifOrientationMode.LeftBottom => AnchorPositionMode.TopLeft,
+                    ExifOrientationMode.BottomRight or ExifOrientationMode.LeftTop => AnchorPositionMode.TopRight,
+                    _ => anchor,
+                },
+                _ => anchor,
+            };
+        }
```
ronaldbarendse, line 180: This should be the inverted switch statement:
```suggestion
          /*
              New anchor position is determined by calculating the direction of the anchor relative to the source image.
              In the following example, the TopRight anchor becomes BottomRight

                          T                              L
              +-----------------------+           +--------------+
              |                      *|           |              |
              |                       |           |              |
            L |           TL          | R         |              |
              |                       |           |              |
              |                       |         B |      LB      | T
              |                       |           |              |
              +-----------------------+           |              |
                          B                       |              |
                                                  |              |
                                                  |             *|
                                                  +--------------+
                                                          R
          */
          => orientation switch
          {
              ExifOrientationMode.TopRight => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Top,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Bottom,
                  AnchorPositionMode.Left => AnchorPositionMode.Right,
                  AnchorPositionMode.Right => AnchorPositionMode.Left,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.TopRight,
                  AnchorPositionMode.TopRight => AnchorPositionMode.TopLeft,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.BottomLeft,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.BottomRight,
                  _ => anchor
              },
              ExifOrientationMode.BottomRight => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Bottom,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Top,
                  AnchorPositionMode.Left => AnchorPositionMode.Right,
                  AnchorPositionMode.Right => AnchorPositionMode.Left,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.BottomRight,
                  AnchorPositionMode.TopRight => AnchorPositionMode.BottomLeft,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.TopLeft,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.TopRight,
                  _ => anchor
              },
              ExifOrientationMode.BottomLeft => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Bottom,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Top,
                  AnchorPositionMode.Left => AnchorPositionMode.Left,
                  AnchorPositionMode.Right => AnchorPositionMode.Right,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.BottomLeft,
                  AnchorPositionMode.TopRight => AnchorPositionMode.BottomRight,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.TopRight,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.TopLeft,
                  _ => anchor
              },
              ExifOrientationMode.LeftTop => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Left,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Right,
                  AnchorPositionMode.Left => AnchorPositionMode.Top,
                  AnchorPositionMode.Right => AnchorPositionMode.Bottom,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.TopLeft,
                  AnchorPositionMode.TopRight => AnchorPositionMode.BottomLeft,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.BottomRight,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.TopRight,
                  _ => anchor
              },
              ExifOrientationMode.RightTop => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Left,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Right,
                  AnchorPositionMode.Left => AnchorPositionMode.Bottom,
                  AnchorPositionMode.Right => AnchorPositionMode.Top,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.BottomLeft,
                  AnchorPositionMode.TopRight => AnchorPositionMode.TopLeft,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.TopRight,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.BottomRight,
                  _ => anchor
              },
              ExifOrientationMode.RightBottom => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Right,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Left,
                  AnchorPositionMode.Left => AnchorPositionMode.Bottom,
                  AnchorPositionMode.Right => AnchorPositionMode.Top,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.BottomRight,
                  AnchorPositionMode.TopRight => AnchorPositionMode.TopRight,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.TopLeft,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.BottomLeft,
                  _ => anchor
              },
              ExifOrientationMode.LeftBottom => anchor switch
              {
                  AnchorPositionMode.Center => AnchorPositionMode.Center,
                  AnchorPositionMode.Top => AnchorPositionMode.Right,
                  AnchorPositionMode.Bottom => AnchorPositionMode.Left,
                  AnchorPositionMode.Left => AnchorPositionMode.Top,
                  AnchorPositionMode.Right => AnchorPositionMode.Bottom,
                  AnchorPositionMode.TopLeft => AnchorPositionMode.TopRight,
                  AnchorPositionMode.TopRight => AnchorPositionMode.BottomRight,
                  AnchorPositionMode.BottomRight => AnchorPositionMode.BottomLeft,
                  AnchorPositionMode.BottomLeft => AnchorPositionMode.TopLeft,
                  _ => anchor
              },
              _ => anchor
          };
```

tests/ImageSharp.Web.Tests/Helpers/ExifOrientationUtilitiesTests.cs
```diff
@@ -0,0 +1,139 @@
...
+                { AnchorPositionMode.BottomRight, ExifOrientationMode.BottomLeft, AnchorPositionMode.TopRight },
+                { AnchorPositionMode.BottomRight, ExifOrientationMode.RightTop, AnchorPositionMode.TopRight },
+                { AnchorPositionMode.BottomRight, ExifOrientationMode.BottomRight, AnchorPositionMode.TopLeft },
+                { AnchorPositionMode.BottomRight, ExifOrientationMode.RightBottom, AnchorPositionMode.TopLeft },
+                { AnchorPositionMode.BottomLeft, ExifOrientationMode.TopRight, AnchorPositionMode.BottomRight },
+                { AnchorPositionMode.BottomLeft, ExifOrientationMode.RightTop, AnchorPositionMode.BottomRight },
+                { AnchorPositionMode.BottomLeft, ExifOrientationMode.BottomLeft, AnchorPositionMode.TopLeft },
+                { AnchorPositionMode.BottomLeft, ExifOrientationMode.LeftBottom, AnchorPositionMode.TopLeft },
+                { AnchorPositionMode.BottomLeft, ExifOrientationMode.BottomRight, AnchorPositionMode.TopRight },
+                { AnchorPositionMode.BottomLeft, ExifOrientationMode.LeftTop, AnchorPositionMode.TopRight },
```
ronaldbarendse, line 127: And this the complete matrix of `AnchorPositionMode` and `ExifOrientationMode` to test:
```suggestion
                { AnchorPositionMode.Center, ExifOrientationMode.Unknown, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.TopLeft, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.TopRight, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.BottomRight, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.BottomLeft, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.LeftTop, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.RightTop, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.RightBottom, AnchorPositionMode.Center },
                { AnchorPositionMode.Center, ExifOrientationMode.LeftBottom, AnchorPositionMode.Center },
                { AnchorPositionMode.Top, ExifOrientationMode.Unknown, AnchorPositionMode.Top },
                { AnchorPositionMode.Top, ExifOrientationMode.TopLeft, AnchorPositionMode.Top },
                { AnchorPositionMode.Top, ExifOrientationMode.TopRight, AnchorPositionMode.Top },
                { AnchorPositionMode.Top, ExifOrientationMode.BottomRight, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Top, ExifOrientationMode.BottomLeft, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Top, ExifOrientationMode.LeftTop, AnchorPositionMode.Left },
                { AnchorPositionMode.Top, ExifOrientationMode.RightTop, AnchorPositionMode.Left },
                { AnchorPositionMode.Top, ExifOrientationMode.RightBottom, AnchorPositionMode.Right },
                { AnchorPositionMode.Top, ExifOrientationMode.LeftBottom, AnchorPositionMode.Right },
                { AnchorPositionMode.Bottom, ExifOrientationMode.Unknown, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Bottom, ExifOrientationMode.TopLeft, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Bottom, ExifOrientationMode.TopRight, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Bottom, ExifOrientationMode.BottomRight, AnchorPositionMode.Top },
                { AnchorPositionMode.Bottom, ExifOrientationMode.BottomLeft, AnchorPositionMode.Top },
                { AnchorPositionMode.Bottom, ExifOrientationMode.LeftTop, AnchorPositionMode.Right },
                { AnchorPositionMode.Bottom, ExifOrientationMode.RightTop, AnchorPositionMode.Right },
                { AnchorPositionMode.Bottom, ExifOrientationMode.RightBottom, AnchorPositionMode.Left },
                { AnchorPositionMode.Bottom, ExifOrientationMode.LeftBottom, AnchorPositionMode.Left },
                { AnchorPositionMode.Left, ExifOrientationMode.Unknown, AnchorPositionMode.Left },
                { AnchorPositionMode.Left, ExifOrientationMode.TopLeft, AnchorPositionMode.Left },
                { AnchorPositionMode.Left, ExifOrientationMode.TopRight, AnchorPositionMode.Right },
                { AnchorPositionMode.Left, ExifOrientationMode.BottomRight, AnchorPositionMode.Right },
                { AnchorPositionMode.Left, ExifOrientationMode.BottomLeft, AnchorPositionMode.Left },
                { AnchorPositionMode.Left, ExifOrientationMode.LeftTop, AnchorPositionMode.Top },
                { AnchorPositionMode.Left, ExifOrientationMode.RightTop, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Left, ExifOrientationMode.RightBottom, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Left, ExifOrientationMode.LeftBottom, AnchorPositionMode.Top },
                { AnchorPositionMode.Right, ExifOrientationMode.Unknown, AnchorPositionMode.Right },
                { AnchorPositionMode.Right, ExifOrientationMode.TopLeft, AnchorPositionMode.Right },
                { AnchorPositionMode.Right, ExifOrientationMode.TopRight, AnchorPositionMode.Left },
                { AnchorPositionMode.Right, ExifOrientationMode.BottomRight, AnchorPositionMode.Left },
                { AnchorPositionMode.Right, ExifOrientationMode.BottomLeft, AnchorPositionMode.Right },
                { AnchorPositionMode.Right, ExifOrientationMode.LeftTop, AnchorPositionMode.Bottom },
                { AnchorPositionMode.Right, ExifOrientationMode.RightTop, AnchorPositionMode.Top },
                { AnchorPositionMode.Right, ExifOrientationMode.RightBottom, AnchorPositionMode.Top },
                { AnchorPositionMode.Right, ExifOrientationMode.LeftBottom, AnchorPositionMode.Bottom },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.Unknown, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.TopLeft, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.TopRight, AnchorPositionMode.TopRight },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.BottomRight, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.BottomLeft, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.LeftTop, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.RightTop, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.RightBottom, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.TopLeft, ExifOrientationMode.LeftBottom, AnchorPositionMode.TopRight },
                { AnchorPositionMode.TopRight, ExifOrientationMode.Unknown, AnchorPositionMode.TopRight },
                { AnchorPositionMode.TopRight, ExifOrientationMode.TopLeft, AnchorPositionMode.TopRight },
                { AnchorPositionMode.TopRight, ExifOrientationMode.TopRight, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.TopRight, ExifOrientationMode.BottomRight, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.TopRight, ExifOrientationMode.BottomLeft, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.TopRight, ExifOrientationMode.LeftTop, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.TopRight, ExifOrientationMode.RightTop, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.TopRight, ExifOrientationMode.RightBottom, AnchorPositionMode.TopRight },
                { AnchorPositionMode.TopRight, ExifOrientationMode.LeftBottom, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.Unknown, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.TopLeft, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.TopRight, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.BottomRight, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.BottomLeft, AnchorPositionMode.TopRight },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.LeftTop, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.RightTop, AnchorPositionMode.TopRight },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.RightBottom, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.BottomRight, ExifOrientationMode.LeftBottom, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.Unknown, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.TopLeft, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.TopRight, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.BottomRight, AnchorPositionMode.TopRight },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.BottomLeft, AnchorPositionMode.TopLeft },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.LeftTop, AnchorPositionMode.TopRight },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.RightTop, AnchorPositionMode.BottomRight },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.RightBottom, AnchorPositionMode.BottomLeft },
                { AnchorPositionMode.BottomLeft, ExifOrientationMode.LeftBottom, AnchorPositionMode.TopLeft },
```

tests/ImageSharp.Web.Tests/Processors/ResizeWebProcessorTests.cs
```diff
@@ -168,7 +168,7 @@ public void ResizeWebProcessor_RespectsOrientation_Center(ushort orientation)
             image.Metadata.ExifProfile.SetValue(ExifTag.Orientation, orientation);
             using var formatted = new FormattedImage(image, PngFormat.Instance);
 
-            PointF expected = GetExpectedCenter(orientation, new PointF(x, y));
+            PointF expected = ExifOrientationUtilities.Transform(new Vector2(x, y), Vector2.Zero, Vector2.One, orientation);
             ResizeOptions options = ResizeWebProcessor.GetResizeOptions(formatted, commands, parser, culture);
             Assert.Equal(expected, options.CenterCoordinates);
```
ronaldbarendse, line 173: This test failed on my machine, because my decimal separator is a comma and the interpolated string above uses the current culture: `$"{x},{y}"`. This can be fixed by updating it to `FormattableString.Invariant($"{x},{y}"))` 👍🏻

tests/ImageSharp.Web.Tests/Helpers/ExifOrientationUtilitiesTests.cs
ronaldbarendse: I applied this test matrix to the current transform and got 2 failing tests:
```
CanTransformAnchor(anchor: TopLeft, orientation: 3, expected: BottomRight)
Assert.Equal() Failure
Expected: BottomRight
Actual:   BottomLeft

CanTransformAnchor(anchor: TopLeft, orientation: 4, expected: BottomLeft)
Assert.Equal() Failure
Expected: BottomLeft
Actual:   BottomRight
```

I believe the test cases are correct, so that would mean the current implementation contains a bug:
- Orientation 3 (bottom/right - upside down) should invert both horizontally and vertically, so top/left should become bottom/right
- Orientation 4 (bottom/left - flipped back-to-front and upside down) should only invert vertically, so top/left should become bottom/left

JimBobSquarePants: There's a good chance you found a bug - orientation is a bit of a brain twister! I will review your matrix and implementation asap. Really appreciate the effort you're making here btw! 👍

tests/ImageSharp.Web.Tests/Helpers/ExifOrientationUtilitiesTests.cs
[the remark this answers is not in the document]
ronaldbarendse: Thinking about this again, clamping the input value should probably be done in the (web)processor itself, just like the `rxy`/`CenterCoordinates` currently is.

The issue I pointed out is still valid though: any known commands are used to generate the cache key/hash, even if the command value is invalid, causing the same resulting image to be cached multiple times using different keys/hashes.

To make this a little bit more complex: different web processors can use the same commands (which might be the case for `width`, `height`, `orient` and `compand`), so command value sanitation isn't something a single processor should be responsible for. You don't want to remove a command that's invalid for one processor, but valid for another (and thus does result in a different result image).

Again, this isn't a big deal and can otherwise be worked around manually by removing any invalid command values using `OnParseCommandsAsync` (or in a more advanced setup, using an HMAC to validate whether the commands/values are generated by your own application).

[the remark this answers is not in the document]
JimBobSquarePants: >The issue I pointed out is still valid though: any known commands are used to generate the cache key/hash, even if the command value is invalid, causing the same resulting image to be cached multiple times using different keys/hashes.

Yeah this is problematic. `OnParseCommandsAsync` is the best place to deal with it for the built in processors. A bounds check there would be sufficient. 

>To make this a little bit more complex: different web processors can use the same commands (which might be the case for `width`, `height`, `orient` and `compand`)

If someone is using width/height and have both `ResizeWebProcessor` and their own custom processor enabled they're in for a whole world of pain.

tests/ImageSharp.Web.Tests/Processors/ResizeWebProcessorTests.cs
JimBobSquarePants: Yeah, my bad.

tests/ImageSharp.Web.Tests/Helpers/ExifOrientationUtilitiesTests.cs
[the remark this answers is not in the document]
JimBobSquarePants: I've added basic sanitation. 

https://github.com/SixLabors/ImageSharp.Web/pull/241/commits/444faa1b558463d134ced64603e56d80cae7abbb

ronaldbarendse:
Great to see this getting merged! 🎉

One note about the sanitization though: this really isn't specific to the `width`, `height` and `rxy` values, as any invalid command value will have this issue. Invalid enums (like `ResizeMode` and `AnchorPositionMode`), booleans or custom command values (like `rsampler`) all return a default when the supplied value is invalid, but still use the input to generate the cache key/hash.

This is basically a chicken-egg problem, as you only know which command values are valid when processing the image (or you have to duplicate all the command parsing), but that's cached by a key that's calculated using values that should prevent all that processing in the first place.

I don't think the solution is in adding custom sanitization to the default `OnParseCommandsAsync` or to introduce some kind of command value validators: a better solution would be to ensure users can't manually fiddle with the command values in the first place, like using the `PresetOnlyQueryCollectionRequestParser` or adding HMAC validation to the generated URL.

The current default sanitization also doesn't remove invalid width or height commands and only removes those commands if both are larger than 4000 (making `?width=4000&height=999999` still an accepted size):

https://github.com/SixLabors/ImageSharp.Web/blob/444faa1b558463d134ced64603e56d80cae7abbb/src/ImageSharp.Web/Middleware/ImageSharpMiddlewareOptions.cs#L27-L49

JimBobSquarePants:
It’s not designed as a catch all. It’s very basic example sanitation to show where this stuff should happen. If someone wants something more comprehensive like HMAC this is where they’d do it. 

Remember I provide tools, not implementations.

ronaldbarendse:
@JimBobSquarePants Looks like you forgot to make `SimpleCommandConverter<T>` public:
https://github.com/SixLabors/ImageSharp.Web/blob/40a50756bca09c43814519e13543bc027c8fb793/src/ImageSharp.Web/Commands/Converters/SimpleCommandConverter%7BT%7D.cs#L14

This class is [required to test the `CropWebProcessor`](https://github.com/umbraco/Umbraco-CMS/blob/e7f74dcb21732aad7c97652125fc633815483c70/tests/Umbraco.Tests.UnitTests/Umbraco.Web.Common/ImageProcessors/CropWebProcessorTests.cs#L46-L53) that ships with Umbraco.

PR #243, opened 2022-04-02
Add FileProviderImageProvider and WebRootImageProvider 

ronaldbarendse:
### Prerequisites

### Description
This adds a `FileProviderImageProvider` and `WebRootImageProvider`, following the pattern mentioned in https://github.com/umbraco/Umbraco-CMS/pull/12185#discussion_r838138281. This is basically a simplified version of PR https://github.com/SixLabors/ImageSharp.Web/pull/207, but these `IImageProvider`s are now included as additional, non-default implementations.

This would allow users to easily switch from the default `PhysicalFileSystemProvider` to the `WebRootImageProvider` by using:
```c#
services.AddImageSharp()
    .RemoveProvider<PhysicalFileSystemProvider>() // or .ClearProviders()
    .AddProvider<WebRootImageProvider>();
```

### DIFF
--- a/src/ImageSharp.Web/Resolvers/FileProviderImageResolver.cs
+++ b/src/ImageSharp.Web/Resolvers/FileProviderImageResolver.cs
@@ -1,0 +1,29 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Extensions.FileProviders;
+
+namespace SixLabors.ImageSharp.Web.Resolvers
+{
+    /// <summary>
+    /// Provides means to manage image buffers from an <see cref="IFileInfo"/> instance.
+    /// </summary>
+    public sealed class FileProviderImageResolver : IImageResolver
+    {
+        private readonly IFileInfo fileInfo;
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


src/ImageSharp.Web/Resolvers/FileProviderImageResolver.cs
```diff
@@ -0,0 +1,29 @@
...
+    /// </summary>
+    public sealed class FileProviderImageResolver : IImageResolver
+    {
+        private readonly IFileInfo fileInfo;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="FileProviderImageResolver"/> class.
+        /// </summary>
+        /// <param name="fileInfo">The file info.</param>
+        public FileProviderImageResolver(IFileInfo fileInfo) => this.fileInfo = fileInfo;
```
ronaldbarendse, line 21: Should this also do a null check?
```suggestion
        public FileProviderImageResolver(IFileInfo fileInfo) => this.fileInfo = fileInfo ?? throw new ArgumentNullException(nameof(fileInfo));;
```

JimBobSquarePants: I wouldn't bother. The default implementation is smart enough to only call this type when the `IFileInfo ` is not null. If someone doesn't do this with their custom code then it's up to them.

JimBobSquarePants (approved):

PR #244, opened 2022-04-02
ImageMetadata and ImageCacheMetadata initialization improvements

ronaldbarendse:
### Prerequisites

### Description
I noticed the `PhysicalFileSystemProvider` always initialized an `ImageMetadata` that requires fetching the `LastWriteTimeUtc` and `Length`, potentially requiring disk I/O (see [reference source](https://github.com/microsoft/referencesource/blob/master/mscorlib/system/io/filesysteminfo.cs#L252-L281)).

I also noticed `PhysicalFileSystemCacheResolver` assumed the `ImageCacheMetadata` was already initialized by `GetMetaDataAsync()`, meaning `OpenReadAsync()` would only work when `GetMetaDataAsync()` was called first. In addition, calling `GetMetaDataAsync()` multiple times would re-initialize the metadata.

### DIFF
--- a/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
+++ b/src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
@@ -70,8 +70,7 @@ namespace SixLabors.ImageSharp.Web.Providers
                 var fileInfo = new FileInfo(fullPath);
                 if (fileInfo.Exists)
                 {
-                    var metadata = new ImageMetadata(fileInfo.LastWriteTimeUtc, fileInfo.Length);
-                    return Task.FromResult<IImageResolver>(new PhysicalFileSystemResolver(fileInfo, metadata));
+                    return Task.FromResult<IImageResolver>(new PhysicalFileSystemResolver(fileInfo));
                 }
             }
 
--- a/src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
+++ b/src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
@@ -14,7 +14,7 @@ namespace SixLabors.ImageSharp.Web.Resolvers
     {
         private readonly FileInfo metaFileInfo;
         private readonly FormatUtilities formatUtilities;
-        private ImageCacheMetadata metadata;
+        private ImageCacheMetadata? metadata;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PhysicalFileSystemCacheResolver"/> class.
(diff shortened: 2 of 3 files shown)


src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
```diff
@@ -30,21 +30,23 @@ public PhysicalFileSystemCacheResolver(FileInfo metaFileInfo, FormatUtilities fo
...
+            return OpenFileStream(path);
+        }
+
+        private async Task<ImageCacheMetadata> ReadMetadataAsync()
+        {
+            using Stream stream = OpenFileStream(this.metaFileInfo.FullName);
+            return await ImageCacheMetadata.ReadAsync(stream);
         }
 
         private static Stream OpenFileStream(string path)
```
JimBobSquarePants, line 52: We should add aggressive inlining here also and make the result return a `FileStream`.

JimBobSquarePants:
I'm not sure about this. The two get methods are only called one time unless something catastrophic and has happened and the request is replayed in which case you want to reload the result. 

Your refactor has added an additional null check to every single request to the cache resolver. I'll add comments with suggestions to mitigate the new overhead but this does not guarantee acceptance.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -70,8 +70,7 @@ public Task<IImageResolver> GetAsync(HttpContext context)
                 var fileInfo = new FileInfo(fullPath);
                 if (fileInfo.Exists)
                 {
-                    var metadata = new ImageMetadata(fileInfo.LastWriteTimeUtc, fileInfo.Length);
```
JimBobSquarePants, line 73: It's actually ok to create the metadata here. `IImageProvider.GetAsync(HttpContext context)` is called exactly one time and the type is only initialized if the source file exists. By doing it like this we avoid the cost of the struct creation per per request since it's a readonly struct.

src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
```diff
@@ -30,21 +30,23 @@ public PhysicalFileSystemCacheResolver(FileInfo metaFileInfo, FormatUtilities fo
         }
 
         /// <inheritdoc/>
-        public async Task<ImageCacheMetadata> GetMetaDataAsync()
-        {
-            using Stream stream = OpenFileStream(this.metaFileInfo.FullName);
-            this.metadata = await ImageCacheMetadata.ReadAsync(stream);
-            return this.metadata;
-        }
+        public async Task<ImageCacheMetadata> GetMetaDataAsync() => this.metadata ??= await this.ReadMetadataAsync();
```
JimBobSquarePants, line 33: This is where potential the issue lies and the rest of the types should have been left untouched. We need to introduce a null check as you have but also try to reduce the overhead of that new check.

src/ImageSharp.Web/Resolvers/PhysicalFileSystemResolver.cs
```diff
@@ -12,21 +12,15 @@ namespace SixLabors.ImageSharp.Web.Resolvers
...
     {
         private readonly FileInfo fileInfo;
-        private readonly ImageMetadata metadata;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="PhysicalFileSystemResolver"/> class.
         /// </summary>
         /// <param name="fileInfo">The input file info.</param>
-        /// <param name="metadata">The image metadata associated with this file.</param>
-        public PhysicalFileSystemResolver(FileInfo fileInfo, in ImageMetadata metadata)
```
JimBobSquarePants, line 22: See how the struct was passed via `in` here? We're passing a ref to the readonly struct and avoiding recreating it each time.

JimBobSquarePants:
I've created #246 to replace this. It means we get to use more of you code and simplify things further. 😉

PR #245, opened 2022-04-02
Remove no-op ImageSharpConfiguration and setup action

ronaldbarendse:
### Prerequisites

### Description
I noticed a no-op `ImageSharpConfiguration` instance and an empty lambda for the setup action was registered: both don't add anything besides additional logic that's run when getting an `IOptions<ImageSharpMiddlewareOptions>`.


src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -121,7 +95,19 @@ public static IImageSharpBuilder AddImageSharp(this IServiceCollection services)
             builder.AddConverter<ColorConverter>();
             builder.AddConverter<EnumConverter>();
 
-            builder.Services.AddSingleton<CommandParser>();
+            return builder;
+        }
+
+        private static IImageSharpBuilder AddDefaultServices(this IImageSharpBuilder builder)
```
ronaldbarendse, line 101: Although default implementations are also set for `IRequestParser`, `IImageCache`, `ICacheKey` and `ICacheHash`, those are actually required for `ImageSharpMiddleware` to activate/resolve from the service container and can easily be replaced with the respective set-methods.

Maybe this extension method should be made public and the `AddImageSharp()` shouldn't call it by default, so you can opt-in to adding these default providers/processors? And following this pattern, maybe even split it into `AddDefaultProviders()` and `AddDefaultProcessors()`?

samples/ImageSharp.Web.Sample/Startup.cs
```diff
@@ -37,71 +37,32 @@ public class Startup
...
-            // Or we can fine-grain control adding the default options and configure all other services.
-            //
-            // this.ConfigureCustomServicesAndDefaultOptions(services);
-
-            // Or we can fine-grain control adding custom options and configure all other services
-            // There are also factory methods for each builder that will allow building from configuration files.
-            //
-            // this.ConfigureCustomServicesAndCustomOptions(services);
-        }
+            => services.AddImageSharp(); // Add the default service and options
```
ronaldbarendse, line 40: The previous code duplicated the defaults that were already set in `AddImageSharp()` and is already documented on https://docs.sixlabors.com/articles/imagesharp.web/gettingstarted.html. Maybe the examples below can also be cleaned up?

JimBobSquarePants: They’re for example purpose only. People don’t always read docs.

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
ronaldbarendse: This would also solve the issue I described in PR https://github.com/SixLabors/ImageSharp.Web/pull/185: the order of registering `IImageProvider`s is important, but can't be (easily) changed. Having a separate method to add the default provider after your own will remove the need to remove and re-add them...

JimBobSquarePants: I don’t want to change registration methods at this time.

JimBobSquarePants: The whole point of the default implementation is that out of the box it works.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilder.cs
```diff
@@ -14,7 +15,7 @@ internal class ImageSharpBuilder : IImageSharpBuilder
         /// Initializes a new instance of the <see cref="ImageSharpBuilder"/> class.
         /// </summary>
         /// <param name="services">The <see cref="IServiceCollection" /> to add services to.</param>
-        public ImageSharpBuilder(IServiceCollection services) => this.Services = services;
+        public ImageSharpBuilder(IServiceCollection services) => this.Services = services ?? throw new ArgumentNullException(nameof(services));
```
JimBobSquarePants, line 18: If `IServiceCollection` is ever null you have bigger problems.

JimBobSquarePants (changes requested):
Please revert all other changes and remove the no-op ImageSharpConfiguration class only. 

The samples serve two purposes and should remain untouched. 
- Education
- API sense checking.

samples/ImageSharp.Web.Sample/Startup.cs
ronaldbarendse: OK, I'll leave the examples in the `ConfigureDefaultServicesAndCustomOptions`, `ConfigureCustomServicesAndDefaultOptions` and `ConfigureCustomServicesAndCustomOptions` methods 👍🏻

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilder.cs
ronaldbarendse: Right, I'll remove this check again 😅

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
ronaldbarendse: I wasn't sure about this either and agree that it's better to have everything working OOTB 😄 I see you've also made inserting providers easier in https://github.com/SixLabors/ImageSharp.Web/pull/247, so this is now less of an issue...

samples/ImageSharp.Web.Sample/Startup.cs
```diff
@@ -37,71 +37,32 @@ public class Startup
         /// </summary>
         /// <param name="services">The collection of service desscriptors.</param>
         public void ConfigureServices(IServiceCollection services)
-        {
```
JimBobSquarePants, line 40: You still need to revert all the changes bar removing `ImageSharpConfiguration`

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -26,55 +24,26 @@ public static class ServiceCollectionExtensions
...
-
-            return builder;
-        }
-
-        private static void AddDefaultServices(
-            IImageSharpBuilder builder,
-            Action<ImageSharpMiddlewareOptions> setupAction)
+        private static IImageSharpBuilder AddDefaultServices(this IImageSharpBuilder builder)
         {
-            builder.Services.Configure(setupAction);
```
ronaldbarendse, line 57: One thing this call did was internally call [`AddOptions()`](https://github.com/dotnet/runtime/blob/bf0492f520c2530c4cc475849e74bf55541e52b8/src/libraries/Microsoft.Extensions.Options/src/OptionsServiceCollectionExtensions.cs#L22-L30), so the generic options interfaces are registered and can be resolved from the service container.

Looking at other extension methods in [ASP.NET Core](https://github.com/dotnet/aspnetcore), they do explicitly call `AddOptions()` as well, so let's make sure ImageSharp.Web does the same (even if it's _probably_ already added anyway, we can't be sure though).

ronaldbarendse: Fixed in https://github.com/SixLabors/ImageSharp.Web/pull/245/commits/fbb6cef1eb6affc8ad470aea252c8f459e886536:
https://github.com/SixLabors/ImageSharp.Web/blob/fbb6cef1eb6affc8ad470aea252c8f459e886536/src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs#L40

samples/ImageSharp.Web.Sample/Startup.cs
JimBobSquarePants: It makes it much more difficult for me to play around with though. We have unit tests for DI.

src/ImageSharp.Web/DependencyInjection/ServiceCollectionExtensions.cs
JimBobSquarePants: I'm sorry but I simply do not follow. Why are you changing any of this? We're only supposed to be removing the no-op `ImageSharpConfiguration`

JimBobSquarePants (approved):
Thanks! 👍

PR #246, opened 2022-04-04
PhysicalFileSystem optimizations

JimBobSquarePants:
### Prerequisites

### Description

Replaces #244 

This refactors the PhysicalFileSystemProvider to use the new `FileProviderImageProvider` base class. In addition it fixes an issue in the `PhysicalFileSystemCacheResolver` to ensure the metadata filestream can only be ever called one time.

CC @ronaldbarendse


src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
```diff
@@ -30,20 +30,23 @@ public class FileProviderImageProvider : IImageProvider
...
+            Guard.NotNull(fileProvider, nameof(fileProvider));
+            Guard.NotNull(formatUtilities, nameof(formatUtilities));
+
+            this.fileProvider = fileProvider;
+            this.formatUtilities = formatUtilities;
         }
 
         /// <inheritdoc/>
-        public ProcessingBehavior ProcessingBehavior { get; protected set; } = ProcessingBehavior.CommandOnly;
+        public virtual ProcessingBehavior ProcessingBehavior { get; protected set; } = ProcessingBehavior.CommandOnly;
```
ronaldbarendse, line 43: Any reason for making this `virtual` _and_ adding the protected setter? This will add the requirement to include the setter if you're overriding this property...

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -41,53 +28,23 @@ public class PhysicalFileSystemProvider : IImageProvider
...
-                if (fileInfo.Exists)
-                {
-                    var metadata = new ImageMetadata(fileInfo.LastWriteTimeUtc, fileInfo.Length);
-                    return Task.FromResult<IImageResolver>(new PhysicalFileSystemResolver(fileInfo, metadata));
-                }
-            }
-
-            return Task.FromResult<IImageResolver>(null);
-        }
+        public override ProcessingBehavior ProcessingBehavior { get; protected set; } = ProcessingBehavior.All;
```
ronaldbarendse, line 36: Why has this changed to `ProcessingBehavior.All`?
```suggestion
        public override ProcessingBehavior ProcessingBehavior { get; protected set; } = ProcessingBehavior.CommandOnly;
```

src/ImageSharp.Web/Resolvers/PhysicalFileSystemCacheResolver.cs
```diff
@@ -32,8 +33,12 @@ public PhysicalFileSystemCacheResolver(FileInfo metaFileInfo, FormatUtilities fo
         /// <inheritdoc/>
         public async Task<ImageCacheMetadata> GetMetaDataAsync()
         {
-            using Stream stream = OpenFileStream(this.metaFileInfo.FullName);
-            this.metadata = await ImageCacheMetadata.ReadAsync(stream);
+            if (this.metadata == default)
+            {
+                using Stream stream = OpenFileStream(this.metaFileInfo.FullName);
```
ronaldbarendse, line 38: This can also use the specific type instead:
```suggestion
                using FileStream stream = OpenFileStream(this.metaFileInfo.FullName);
```

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
```diff
@@ -101,5 +58,18 @@ internal static string GetProviderRoot(PhysicalFileSystemProviderOptions provide
...
+            PhysicalFileSystemProviderOptions options,
+#if NETCOREAPP2_1
+            IHostingEnvironment environment)
+#else
+            IWebHostEnvironment environment)
+#endif
+        {
+            Guard.NotNull(options, nameof(options));
+            Guard.NotNull(environment, nameof(environment));
+            return new(GetProviderRoot(options, environment.WebRootPath, environment.ContentRootPath));
```
ronaldbarendse, line 72: Keep in mind that the `PhysicalFileProvider` requires the directory to exist (it throws an exception if not). I guess that's fine, because you don't want to create directories in constructors (would be a nightmare when used in tests) and this provider is otherwise useless anyway...

If you have a custom provider root set (to a fully qualified path), you can also use the configure callback to ensure the directory exists before the constructor is called:

```c#
.Configure<PhysicalFileSystemProviderOptions>(options =>
{
    if (!string.IsNullOrEmpty(options.ProviderRootPath))
    {
        Directory.CreateDirectory(options.ProviderRootPath);
    }
})
```

ronaldbarendse (approved):

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: Because the image could be stored in a folder that the static file middleware cannot touch. I’m still undecided about that and am thinking of making the setter public instead so users can use a factory to set it up

JimBobSquarePants: Or even remove the setter and add the property to the options. That would be better actually…

src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
JimBobSquarePants: We need to be able to make this property configurable.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
JimBobSquarePants: I've reverted the default since we default to wwwroot. Instead the value is passed via the options. We still require the protected set in order to set that value though unless we pass it via the constructor.

src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
```diff
@@ -29,21 +29,26 @@ public class FileProviderImageProvider : IImageProvider
         /// Initializes a new instance of the <see cref="FileProviderImageProvider"/> class.
         /// </summary>
         /// <param name="fileProvider">The file provider.</param>
+        /// <param name="processingBehavior">The processing behavior.</param>
         /// <param name="formatUtilities">Contains various format helper methods based on the current configuration.</param>
-        public FileProviderImageProvider(IFileProvider fileProvider, FormatUtilities formatUtilities)
+        protected FileProviderImageProvider(IFileProvider fileProvider, ProcessingBehavior processingBehavior, FormatUtilities formatUtilities)
```
ronaldbarendse, line 34: Can the `processingBehavior` parameter move to the end to align with the field/property setting order in the constructor (and also update the XML documentation order):
```suggestion
        protected FileProviderImageProvider(IFileProvider fileProvider, FormatUtilities formatUtilities, ProcessingBehavior processingBehavior)
```

And optionally add a default value?
```suggestion
        protected FileProviderImageProvider(IFileProvider fileProvider, FormatUtilities formatUtilities, ProcessingBehavior processingBehavior = ProcessingBehavior.CommandOnly)
```

src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
```diff
@@ -29,21 +29,26 @@ public class FileProviderImageProvider : IImageProvider
...
+            Guard.NotNull(formatUtilities, nameof(formatUtilities));
+
+            this.fileProvider = fileProvider;
+            this.formatUtilities = formatUtilities;
+            this.ProcessingBehavior = processingBehavior;
         }
 
         /// <inheritdoc/>
-        public ProcessingBehavior ProcessingBehavior { get; protected set; } = ProcessingBehavior.CommandOnly;
+        public ProcessingBehavior ProcessingBehavior { get; }
```
ronaldbarendse, line 45: Or remove the `processingBehavior` parameter from the constructor and make the property virtual again (aligning it with the `Match` property):

```suggestion
        public virtual ProcessingBehavior ProcessingBehavior { get; } = ProcessingBehavior.CommandOnly;
```

The `PhysicalFileSystemProvider` implmentation can then override this property to be able to set the value in the constructor and `WebRootImageProvider` can simply use the default value (without having to specify it when calling the base constructor).

My reasoning for this is that base/abstract class shouldn't be influenced by derived classes, so you should add constructor parameters because one single derived class want to set this property.

ronaldbarendse (changes requested):

src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
JimBobSquarePants: We don't do default values. They make tracking API changes more difficult.

src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
JimBobSquarePants: >My reasoning for this is that base/abstract class shouldn't be influenced by derived classes, so you should add constructor parameters because one single derived class want to set this property.

This doesn't make any sense. Base class implementations absolutely are influenced by the parameters passed. 

I'm adding the constructor parameter to ensure that the API shape matches the interface. `ProcessingBehavior` is a readonly required field, therefor it should be passed via the constructor.

src/ImageSharp.Web/Providers/PhysicalFileSystemProvider.cs
ronaldbarendse: Having it configurable for the `PhysicalFileSystemProvider` does indeed make sense if you've specified a provider root path that isn't inside the wwwroot (and therefore not served by the static file middleware).

Because the `WebRootImageProvider` is tightly coupled with the `WebRootFileProvider`, having that always use `ProcessingBehavior.CommandOnly` also makes sense 👍🏻

ronaldbarendse (approved):

PR #247, opened 2022-04-04
Add InsertProvider extensions and tests

JimBobSquarePants:
### Prerequisites

### Description

Adds the ability to insert an `IImageProvider` or factory into the provider collection. This reduces boilerplate code when adding multiple providers.


ronaldbarendse (approved):
Looks good, but the lifetime check can probably be skipped (same for removing), because clearing the providers/processors uses `Services.RemoveAll()` that only looks at the service type anyway.

JimBobSquarePants:
@ronaldbarendse Agreed. Removed that additional check.

PR #248, opened 2022-04-04
Improve default OnParseCommandsAsync sanitization

ronaldbarendse:
### Prerequisites

### Description
This improves the default `OnParseCommandsAsync` sanitization by removing an invalid `width` or `height` command value, which would be <= 0 or > 4000.

As mentioned in https://github.com/SixLabors/ImageSharp.Web/pull/241#issuecomment-1088018691, this removes either command if the value is invalid, so `?width=4000&height=999999` will only resize the width to 4000 now.

I've also removed the `rxy` sanitization, as that kind of validation is an entirely separate issue...


JimBobSquarePants:
Why is it a separate issue? I really which you would use the discussion channel first

src/ImageSharp.Web/Middleware/ImageSharpMiddlewareOptions.cs
```diff
@@ -24,28 +24,17 @@ public class ImageSharpMiddlewareOptions
                 return Task.CompletedTask;
             }
 
-            // It's a good idea to have this to provide very basic security.
```
JimBobSquarePants, line 27: I'm gonna scrap all of this and introduce optional HMAC functionality instead.

ronaldbarendse: Although I would love to see built-in HMAC validation support, that would be a major feature and something that would only work if it's fully integrated in the consuming project (all image processing URLs would need to be generated with logic that adds the HMAC hash/token).

This is also something that Orchard Core has [already implemented themselves](https://github.com/OrchardCMS/OrchardCore/blob/c8b2954e038aaa54963b60bc6cfe193cfea54799/src/OrchardCore.Modules/OrchardCore.Media/Processing/MediaImageSharpConfiguration.cs#L36-L63) (maybe check with @deanmarcussen).

Having some default validation on the maximum width/height actually makes sense, as values >4000 are still valid, but would easily drain your CPU, memory and cache/disk resources.

JimBobSquarePants:
I'm closing this since #250 removes any default sanitation and I want users to use that over any custom parsing.

PR #249, opened 2022-04-05
Make InsertProvider idempotent and remove reflection

ronaldbarendse:
### Prerequisites

### Description
As mentioned in https://github.com/SixLabors/ImageSharp.Web/pull/247#issuecomment-1088626468, `InsertProvider` doesn't behave as expected and isn't idempotent: inserting a provider to a specific index only changed it to that index if it wasn't already added at a lower index:

```c#
builder.ClearProviders();
builder.AddProvider<PhysicalFileSystemProvider>();
builder.InsertProvider<MockImageProvider>(0); // Adds it before PhysicalFileSystemProvider (correct)
builder.InsertProvider<MockImageProvider>(1); // Doesn't update it to index 1 (incorrect)

builder.ClearProviders();
builder.AddProvider<PhysicalFileSystemProvider>();
builder.InsertProvider<MockImageProvider>(1); // Adds it after PhysicalFileSystemProvider (correct)
builder.InsertProvider<MockImageProvider>(0); // Inserts it before PhysicalFileSystemProvider (correct)
```

This PR also renames the generic type argument to `TImplementation`, because the some names were hard to distinguish with the interface name (e.g. `TCacheHash` and `ICacheHash`) and this aligns with the naming used on the service collection extensions.

-----------------------

I've also added a `PostConfigure()` extension method, so you can easily run code after the final state of the options instance is processed. This will allow users to e.g. easily ensure the directory used by the `PhysicalFileSystemProvider` is always created (as mentioned in https://github.com/SixLabors/ImageSharp.Web/pull/246#discussion_r842179475):

```c#
services.AddImageSharp().PostConfigure<PhysicalFileSystemProviderOptions>(options =>
{
    if (!string.IsNullOrEmpty(options.ProviderRootPath))
    {
        Directory.CreateDirectory(options.ProviderRootPath);
    }
})
```


src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -377,7 +371,28 @@ public static IImageSharpBuilder Configure<TOptions>(this IImageSharpBuilder bui
...
+        {
+            builder.Services.PostConfigure(configureOptions);
+
             return builder;
         }
+
+        private static Type GetImplementationType(this ServiceDescriptor descriptor)
+            => descriptor.ImplementationType
+            ?? descriptor.ImplementationInstance?.GetType()
+            ?? descriptor.ImplementationFactory?.GetType().GenericTypeArguments[1];
```
ronaldbarendse, line 396: This is inspired by the [internal `ServiceDescriptor.GetImplementationType()` method](https://github.com/dotnet/runtime/blob/34f2b587f3512d3974d73b316e565194df34eea3/src/libraries/Microsoft.Extensions.DependencyInjection.Abstractions/src/ServiceDescriptor.cs#L95-L116).

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -23,14 +22,14 @@ public static class ImageSharpBuilderExtensions
         /// <summary>
         /// Sets the given <see cref="IRequestParser"/> adding it to the service collection.
         /// </summary>
-        /// <typeparam name="TParser">The type of class implementing <see cref="IRequestParser"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IRequestParser"/> to add.</typeparam>
```
JimBobSquarePants, line 25: `TParser` is better here. Even `T` is better than `TImplementation` which is wordy and doesn't add value.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -42,22 +41,22 @@ public static IImageSharpBuilder SetRequestParser<TParser>(this IImageSharpBuild
...
         /// Sets the given <see cref="IImageCache"/> adding it to the service collection.
         /// </summary>
-        /// <typeparam name="TCache">The type of class implementing <see cref="IImageCache"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageCache"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder SetCache<TCache>(this IImageSharpBuilder builder)
-            where TCache : class, IImageCache
+        public static IImageSharpBuilder SetCache<TImplementation>(this IImageSharpBuilder builder)
+            where TImplementation : class, IImageCache
```
JimBobSquarePants, line 56: Same as above. The naming was explicit and terse.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -23,14 +22,14 @@ public static class ImageSharpBuilderExtensions
...
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder SetRequestParser<TParser>(this IImageSharpBuilder builder)
-            where TParser : class, IRequestParser
+        public static IImageSharpBuilder SetRequestParser<TImplementation>(this IImageSharpBuilder builder)
+            where TImplementation : class, IRequestParser
         {
-            var descriptor = new ServiceDescriptor(typeof(IRequestParser), typeof(TParser), ServiceLifetime.Singleton);
-            builder.Services.Replace(descriptor);
+            builder.Services.Replace(ServiceDescriptor.Singleton<IRequestParser, TImplementation>());
```
JimBobSquarePants, line 31: Much neater

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -69,22 +68,22 @@ public static IImageSharpBuilder SetCache<TCache>(this IImageSharpBuilder builde
...
         /// Sets the given <see cref="ICacheKey"/> adding it to the service collection.
         /// </summary>
-        /// <typeparam name="T">The type of class implementing <see cref="ICacheKey"/> to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="ICacheKey"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder SetCacheKey<T>(this IImageSharpBuilder builder)
-            where T : class, ICacheKey
+        public static IImageSharpBuilder SetCacheKey<TImplementation>(this IImageSharpBuilder builder)
+            where TImplementation : class, ICacheKey
```
JimBobSquarePants, line 83: `TCacheKey` would be better

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -96,22 +95,22 @@ public static IImageSharpBuilder SetCacheKey<T>(this IImageSharpBuilder builder)
...
 
         /// <summary>
         /// Sets the given <see cref="ICacheHash"/> adding it to the service collection.
         /// </summary>
-        /// <typeparam name="TCacheHash">The type of class implementing <see cref="ICacheHash"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="ICacheHash"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilder builder)
-            where TCacheHash : class, ICacheHash
```
JimBobSquarePants, line 111: `TCacheHash `

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -123,91 +122,91 @@ public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilde
...
         /// Adds the given <see cref="IImageProvider"/> to the provider collection within the service collection.
         /// </summary>
-        /// <typeparam name="TProvider">The type of class implementing <see cref="IImageProvider"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageProvider"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder AddProvider<TProvider>(this IImageSharpBuilder builder)
-            where TProvider : class, IImageProvider
+        public static IImageSharpBuilder AddProvider<TImplementation>(this IImageSharpBuilder builder)
+            where TImplementation : class, IImageProvider
```
JimBobSquarePants, line 137: `TProvider`

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -123,91 +122,91 @@ public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilde
...
         /// <summary>
         /// Adds the given <see cref="IImageProvider"/> to the provider collection within the service collection.
         /// </summary>
-        /// <typeparam name="TProvider">The type of class implementing <see cref="IImageProvider"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageProvider"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <param name="implementationFactory">The factory method for returning a <see cref="IImageProvider"/>.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder AddProvider<TProvider>(this IImageSharpBuilder builder, Func<IServiceProvider, TProvider> implementationFactory)
-            where TProvider : class, IImageProvider
```
JimBobSquarePants, line 152: `TProvider `

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -123,91 +122,91 @@ public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilde
...
         /// </summary>
-        /// <typeparam name="TProvider">The type of class implementing <see cref="IImageProvider"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageProvider"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <param name="index">The zero-based index at which the provider should be inserted.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder InsertProvider<TProvider>(this IImageSharpBuilder builder, int index)
-            where TProvider : class, IImageProvider
+        public static IImageSharpBuilder InsertProvider<TImplementation>(this IImageSharpBuilder builder, int index)
+            where TImplementation : class, IImageProvider
```
JimBobSquarePants, line 167: `TProvider `

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -123,91 +122,91 @@ public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilde
...
-        /// <typeparam name="TProvider">The type of class implementing <see cref="IImageProvider"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageProvider"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <param name="index">The zero-based index at which the provider should be inserted.</param>
         /// <param name="implementationFactory">The factory method for returning a <see cref="IImageProvider"/>.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder InsertProvider<TProvider>(this IImageSharpBuilder builder, int index, Func<IServiceProvider, TProvider> implementationFactory)
-            where TProvider : class, IImageProvider
+        public static IImageSharpBuilder InsertProvider<TImplementation>(this IImageSharpBuilder builder, int index, Func<IServiceProvider, TImplementation> implementationFactory)
+            where TImplementation : class, IImageProvider
```
JimBobSquarePants, line 188: `TProvider `

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -123,91 +122,91 @@ public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilde
...
         /// Removes the given <see cref="IImageProvider"/> from the provider collection within the service collection.
         /// </summary>
-        /// <typeparam name="TProvider">The type of class implementing <see cref="IImageProvider"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageProvider"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder RemoveProvider<TProvider>(this IImageSharpBuilder builder)
-            where TProvider : class, IImageProvider
+        public static IImageSharpBuilder RemoveProvider<TImplementation>(this IImageSharpBuilder builder)
+            where TImplementation : class, IImageProvider
```
JimBobSquarePants, line 207: `TProvider`

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
JimBobSquarePants: I think at the time I first started writing all this stuff. .NET Core 1 was still a thing and you needed `GetMethodInfo()`. This is much better.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -377,7 +371,28 @@ public static IImageSharpBuilder Configure<TOptions>(this IImageSharpBuilder bui
...
+        }
+
+        /// <summary>
+        /// Registers an action used to configure a particular type of options. Note: These are run after all <see cref="Configure{TOptions}(IImageSharpBuilder, Action{TOptions})"/>.
+        /// </summary>
+        /// <typeparam name="TOptions">The options type to be configured.</typeparam>
+        /// <param name="builder">The core builder.</param>
+        /// <param name="configureOptions">The action used to configure the options.</param>
+        /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
+        public static IImageSharpBuilder PostConfigure<TOptions>(this IImageSharpBuilder builder, Action<TOptions> configureOptions)
```
JimBobSquarePants, line 385: What benefit do you see to this rather than having the explicit `Configure<T>` methods for each type? 

`PostConfigure<T>` is designed to be called from within extension methods to add services to ensure that those options are not overwritten. It doesn't really have a place in the fluent builder API.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -285,53 +281,50 @@ public static IImageSharpBuilder RemoveProcessor<TProcessor>(this IImageSharpBui
...
         /// <summary>
         /// Adds the given <see cref="ICommandConverter"/> to the converter collection within the service collection.
         /// </summary>
-        /// <typeparam name="TConverter">The type of class implementing <see cref="ICommandConverter"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="ICommandConverter"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <param name="implementationFactory">The factory method for returning a <see cref="ICommandConverter"/>.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder AddConverter<TConverter>(this IImageSharpBuilder builder, Func<IServiceProvider, TConverter> implementationFactory)
-            where TConverter : class, ICommandConverter
```
JimBobSquarePants, line 314: `TConverter`

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -223,53 +222,50 @@ public static IImageSharpBuilder RemoveProvider<TProvider>(this IImageSharpBuild
...
 
         /// <summary>
         /// Adds the given <see cref="IImageWebProcessor"/> to the processor collection within the service collection.
         /// </summary>
-        /// <typeparam name="TProcessor">The type of class implementing <see cref="IImageWebProcessor"/>to add.</typeparam>
+        /// <typeparam name="TImplementation">The type of class implementing <see cref="IImageWebProcessor"/> to add.</typeparam>
         /// <param name="builder">The core builder.</param>
         /// <returns>The <see cref="IImageSharpBuilder"/>.</returns>
-        public static IImageSharpBuilder AddProcessor<TProcessor>(this IImageSharpBuilder builder)
-            where TProcessor : class, IImageWebProcessor
```
JimBobSquarePants, line 238: `TProcessor`

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
```diff
@@ -123,91 +122,91 @@ public static IImageSharpBuilder SetCacheHash<TCacheHash>(this IImageSharpBuilde
...
-        public static IImageSharpBuilder InsertProvider<TProvider>(this IImageSharpBuilder builder, int index)
-            where TProvider : class, IImageProvider
+        public static IImageSharpBuilder InsertProvider<TImplementation>(this IImageSharpBuilder builder, int index)
+            where TImplementation : class, IImageProvider
         {
             var descriptors = builder.Services.Where(x => x.ServiceType == typeof(IImageProvider)).ToList();
-            builder.ClearProviders();
-
-            descriptors.Insert(index, ServiceDescriptor.Singleton<IImageProvider, TProvider>());
+            descriptors.RemoveAll(x => x.GetImplementationType() == typeof(TImplementation));
```
JimBobSquarePants, line 170: Ah right, so you're removing any previous instances of the same type. It's really odd that someone would add multiple instances but I'll go with it.

JimBobSquarePants (changes requested):
I'd say revert everything but the actual fix. 

You'd save yourself a LOT of time if you just chatted first before opening a PR.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
ronaldbarendse: I wasn't sure why `Configure<T>()` was added to `IImageSharpBuilder` at first, as it's just a proxy to `IServiceCollection`, but it makes it a lot easier to configure ImageSharp related options in a fluent way (although you can configure any options).

Maybe it's indeed a sensible idea to add overloads for the specific ImageSharp option types, so it easier to discover relevant options, although that would mean the Azure and S3 packages will also need to provide their own overloads for their types. You could also introduce an decorator interface and limit the generic type to e.g. `where TOptions : IImageSharpOptions`, but that would make discovery of the available ImageSharp options a bit harder again. What's your opinion on this?

As for `PostConfigure<T>()`, that basically does the same as `Configure<T>()`, but is invoked last, ensuring you have the final state of the options (in e.g. case multiple configure calls are done).

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
ronaldbarendse: I hesitated between changing it to `T` or `TImplementation` because the latter is indeed a bit wordy and doesn't add the same value as when used in the service collection extension methods (that have `<TService, TImplementation>`).

I've mentioned my reasoning for this in the PR description and would actually prefer to change it to `T` instead, if you're okay with that:

> This PR also renames the generic type argument to TImplementation, because the some names were hard to distinguish with the interface name (e.g. TCacheHash and ICacheHash) and this aligns with the naming used on the service collection extensions.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
ronaldbarendse: Had a quick look and you can't create overloads for `Configure<T>(this IImageSharpBuilder builder, IConfiguration config)` that specify specific ImageSharp options, so the only way to limit the configure methods on `IImageSharpBuilder` is to create a marker interface...

I think only allowing ImageSharp related configuration on the `IImageSharpBuilder` is actually a better API design and you can still (post)configure the options directly on the `IServiceCollection` if you want...

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
ronaldbarendse: Agreed that it might seem odd, but you don't always own all code and know whether other users have already added a provider. And making the API idempotent is a nice bonus 😉

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
JimBobSquarePants: Marker interfaces are a code smell.

Configure is fine as it is, the flexibility is a strength and it works really well when building the configuration sequentially. The examples in the Samples startup are indicative of that strength as does the config in the the unit tests.

src/ImageSharp.Web/DependencyInjection/ImageSharpBuilderExtensions.cs
JimBobSquarePants: The existing naming is a convention we use throughout the Six Labors libraries e.g. TPixel and works well for us. It’s far more consistent than the .NET runtime and I’d like to keep it as is.

JimBobSquarePants (approved):
Thanks for the fix 👍

PR #250, opened 2022-04-05
Add HMAC authentication options.

JimBobSquarePants:
### Prerequisites

### Description

This PR builds upon an idea of @ronaldbarendse to allow adding optional Hash-based Message Authentication Code (HMAC) authentication to image requests. This replaces the previous, flaky sanitation found in `ImageSharpMiddlewareOptions.OnParseCommandsAsync` and allows authentication of requests against public facing web applications.

Two new properties have been added to `ImageSharpMiddlewareOptions`

- `HMACSecretKey` a `byte[]` containing a key used for HMAC hashing
- `OnComputeHMACAsync` a `Func<ImageCommandContext, byte[], Task<string>>` that contains the default HMAC hashing implementation.

In addition a public `HMACUtilities` has been implemented allowing hashing at 256, 384, and 512 bits and a `CaseHandlingUriBuilder` exposed to allow URI generation. 

By default this authentication is turned off but can be enabled by setting the `HMACSecretKey` property to an array of length greater than 0.

Once enabled, the default implementation uses a hash size of  256 bits and creates the hash from an invariant lowercase version of the relative image request uri. 

```
/myimage.jpeg?width=100&v=12345678
```

Hashes should be appended to the uri in the following format.

```
&hmac=HASH
```

 where `HASH` is calculated in the following manner when using the default implementation. 

```
string token= HMACUtilities.ComputeHMACSHA256(uri.ToLowerInvariant(), options.HMACSecretKey);
uri = uri + "&" + HMACUtilities.TokenCommand + "=" + token;
```

The `hmac` command key is a constant found at `HMACUtilities.TokenCommand`.

Token resolution is one of the very first operations performed in the middleware and, as such, authentication failure will result in a <strike>401</strike> 400 response with no further processing.


JimBobSquarePants:
I need to fix the tests here. I've got tests running two separate servers running in parallel hitting the same diskspace

src/ImageSharp.Web/Caching/SHA256CacheHash.cs
```diff
@@ -33,18 +33,16 @@ public SHA256CacheHash(IOptions<ImageSharpMiddlewareOptions> options)
...
-
             byte[] buffer = null;
+
             try
             {
-                buffer = ArrayPool<byte>.Shared.Rent(byteCount);
-                return HashValue(value, length, buffer.AsSpan(0, byteCount));
+                // Allocating a buffer from the pool is ~27% slower than stackalloc so use that for short strings
+                Span<byte> bytes = byteCount <= 128
+                    ? stackalloc byte[byteCount]
```
lahma, line 42: There was a Twitter thread recently about using static value for stackalloc being better, like stackalloc[128].AsSpan(requested).

https://twitter.com/EgorBo/status/1508069816275513344?t=zxpY4AQ_9JddgbLKat0Yrw&s=19

JimBobSquarePants: Ah yeah, then you slice it. I remember that Tweet. Thanks!

lahma: Cool,  I can't remember if SkipLocalsInit could  be just shim for older frameworks via custom attribute and compiler would pick it up. Great work here as always!

JimBobSquarePants: No idea tbh! It's really hard to keep track of what is shimmable. Thanks!

ronaldbarendse:
You might want to look into [ASP.NET Core Data Protection](https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/introduction), so the key configuration/storage/rotation can be handled for you. You can also use different purpose names to support multi-tenancy (so one tenant can't create valid HMACs for other tenants).

I also think returning a [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) or otherwise [400 Bad Request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400) is more appropriate then the [401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401), as that should include a `WWW-Authenticate` header to allow reauthentication.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -211,8 +220,22 @@ private async Task Invoke(HttpContext httpContext, bool retry)
...
 
-            await this.options.OnParseCommandsAsync.Invoke(
-                new ImageCommandContext(httpContext, commands, this.commandParser, this.parserCulture));
+            ImageCommandContext imageCommandContext = new(httpContext, commands, this.commandParser, this.parserCulture);
+
+            if (doHMAC)
+            {
+                // Compare the passed token to our generated mac.
+                string mac = await this.options.OnComputeHMACAsync(imageCommandContext, secret);
+                if (mac != token)
```
deanmarcussen, line 229: Just looking quickly, but isn't this going to apply to every request
Even requests not meant for ImageSharp?

In theory looks good, I did something similar for OC some time ago, but processed it in `OnParseCommandsAsync` (and then @lahma made it faster). Think I quite heavily memory cached the query strings, so we weren't calculating hashes all the time.

I went with HMAC as well, rather than DP.
When using DP, the keys rotate / an encrypted value, may not always be the same hash, which causes issues with browser caching, as the same request, may not look like the same query string anymore, due to a different hash.

So I felt it better to use a fixed key.

We aren't returning an unauthorized, I probably don't see it as that kind of feature - sounds like something for an actual auth handler.

Our approach was just to strip the commands, if the hash was invalid. Which then lets the request be parsed to the next middleware, i.e. static files, so the image is served, but at full size.

Will look more tomorrow when back at keyboard.

JimBobSquarePants: > Just looking quickly, but isn't this going to apply to every request
Even requests not meant for ImageSharp?

That is a VERY good point, will shift it till after I have a match.

I chose to actually reject the call because I want to avoid all further processing overheads.

JimBobSquarePants:
@ronaldbarendse Still not sure about the response code - good point about the header.

Re the libraries you pointed out. Thanks but I don’t really want to use rotation as a fixed key is easier an suitable for our needs and HMAC isn’t hard to implement. The key itself can be stored wherever, it just needs to be assigned at startup.

JimBobSquarePants:
@deanmarcussen I've moved the handling until after we've verified it's a valid image request and now temporarily cache the generated tokens.

@ronaldbarendse I've added the header to the response now. I've also made the token generation case sensitive. This is to avoid users doing `imageUrl.ToLowerInvariant()` in the calling code when generating tokens. I can't decide whether this is wise though (_it shouldn't be an attack vector_) so would value your input.

ronaldbarendse (changes requested):

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -211,8 +226,9 @@ private async Task Invoke(HttpContext httpContext, bool retry)
                 }
             }
 
-            await this.options.OnParseCommandsAsync.Invoke(
-                new ImageCommandContext(httpContext, commands, this.commandParser, this.parserCulture));
+            ImageCommandContext imageCommandContext = new(httpContext, commands, this.commandParser, this.parserCulture);
+
+            await this.options.OnParseCommandsAsync.Invoke(imageCommandContext);
```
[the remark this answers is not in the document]
JimBobSquarePants: Yup. Will fix.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -252,6 +288,13 @@ private async Task Invoke(HttpContext httpContext, bool retry)
                 retry);
         }
 
+        private static void SetUnauthorized(HttpContext httpContext)
+        {
+            httpContext.Response.Clear();
+            httpContext.Response.Headers.Add("WWW-Authenticate", "HMAC realm=\"" + httpContext.Request.Host + "\"");
```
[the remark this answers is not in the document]
JimBobSquarePants: I actually got the same advice from the creator of OpenID-addict. I bow to your superior knowledge.

src/ImageSharp.Web/CaseHandlingUriBuilder.cs
```diff
@@ -6,18 +6,29 @@
...
     {
         private static readonly SpanAction<char, (bool LowerInvariant, string Host, string PathBase, string Path, string Query)> InitializeAbsoluteUriStringSpanAction = new(InitializeAbsoluteUriString);
 
+        /// <summary>
+        /// Provides Uri case handling options.
+        /// </summary>
         public enum CaseHandling
         {
+            /// <summary>
+            /// No adjustments to casing are made.
```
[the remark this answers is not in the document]
JimBobSquarePants: This behaviour describes any alterations to the input value.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -191,9 +197,18 @@ public class ImageSharpMiddleware
 
         private async Task Invoke(HttpContext httpContext, bool retry)
         {
-            // We expect to get concrete collection type which removes virtual dispatch concerns and enumerator allocations
             CommandCollection commands = this.requestParser.ParseRequestCommands(httpContext);
 
+            // First check for a HMAC token and capture before the command is stripped out.
+            byte[] secret = this.options.HMACSecretKey;
```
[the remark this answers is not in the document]
JimBobSquarePants: There's no issue sharing a shared secret across a multitenant install. This is designed to prevent external bad actors not internal resolution of requests. 

You would use a custom `IImageProvider` or authorization middleware that sites in front of this middleware to segregate tenants.

src/ImageSharp.Web/CaseHandlingUriBuilder.cs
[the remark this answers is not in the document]
JimBobSquarePants: I've switched back to lower invariant for the default generation.

PR #256, opened 2022-04-25
Remove HMAC caching by token

ronaldbarendse:
### Prerequisites

### Description
As mentioned in https://github.com/SixLabors/ImageSharp.Web/pull/250/files#r857755086, caching generated HMACs based on the token completely bypasses the validation, because the cache key doesn't change when the commands are changed and therefore returns the same valid HMAC when the token is used with different commands.

I've thought about creating a better key for the HMAC cache, but that would also require additional processing and you might get similar issues when that cache key doesn't use the same information as `OnComputeHMACAsync` does (e.g. the HMAC cache key uses a relative URL, but the HMAC is computed using an absolute URL).

The overhead of generating HMACs is also relatively low and attackers could otherwise just change the URL slightly to bypass the cached HMACs anyway. This is why I think completely removing the HMAC cache seems like the most reasonable fix.

-------------------

You can quickly test this by using the following code and appending `&hmac=` with the logged value:
```c#
using System.Security.Cryptography;
using Microsoft.AspNetCore.Http.Extensions;

services.AddImageSharp(options =>
{
    // Randomly generate a HMAC secret key
    byte[] secret = new byte[64];
    RandomNumberGenerator.Create().GetBytes(secret);
    options.HMACSecretKey = secret;

    // Log computed HMAC to console for testing
    var onComputeHMACAsync = options.OnComputeHMACAsync;
    options.OnComputeHMACAsync = async (context, secret) =>
    {
        var hmac = await onComputeHMACAsync(context, secret);

        Console.WriteLine($"Computed HMAC for {context.Context.Request.GetDisplayUrl()}: {hmac}");

        return hmac;
    };
});

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -42,13 +42,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
             = new(1024, TimeSpan.FromSeconds(30));
 
         /// <summary>
-        /// Used to temporarily store cached HMAC-s to reduce the overhead of HMAC token generation.
-        /// </summary>
-        private static readonly ConcurrentTLruCache<string, string> HMACTokenLru
-            = new(1024, TimeSpan.FromSeconds(30));
-
-        /// <summary>
-        /// The function processing the Http request.
+        /// The function processing the next HTTP request.
         /// </summary>
         private readonly RequestDelegate next;
 
@@ -201,11 +195,11 @@ namespace SixLabors.ImageSharp.Web.Middleware
 
             // First check for a HMAC token and capture before the command is stripped out.
             byte[] secret = this.options.HMACSecretKey;
-            bool checkHMAC = false;
+            bool computeHMAC = false;
             string token = null;
             if (secret?.Length > 0)
             {
-                checkHMAC = true;
+                computeHMAC = true;
                 token = commands.GetValueOrDefault(HMACUtilities.TokenCommand);
             }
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by JimBobSquarePants** [comment]
What are you on about? 

Can you please raise an issue proving a bug in the implementation. There is protocol here and you’re not following it and adding unnecessary work for me chasing down random comments littered though the development history.

**Comment by ronaldbarendse** [comment]
> What are you on about?

I thought this was just too obvious if you look at this single line of code:

https://github.com/SixLabors/ImageSharp.Web/blob/026b776a2fce70182053e49e819612bd0d953231/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs#L259

The HMAC is generated based on the information in the `ImageCommandContext` (by default the relative path and parsed commands), but then cached using only the token (a single command from the request). Any subsequent request using the same token will therefore return the same valid HMAC from the cache, even if you've supplied different commands in that request!

**Comment by JimBobSquarePants** [comment]
Wouldn't the fix then be as simple as using the full encoded (_or even display_) URL as the key?

File: src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs [hunk 1]
```diff
@@ -201,11 +195,11 @@ private async Task Invoke(HttpContext httpContext, bool retry)
 
             // First check for a HMAC token and capture before the command is stripped out.
             byte[] secret = this.options.HMACSecretKey;
-            bool checkHMAC = false;
+            bool computeHMAC = false;
```
JimBobSquarePants, line 198: Changing the variable name here incorrectly changes the meaning of the bool field.

ronaldbarendse: That's because the meaning _has_ changed! Instead of always requiring a valid token when a secret is configured, it's dependent on whether the `OnComputeHMACAsync` returns null or not.

Together with removing the caching, this allows you to bypass the HMAC check based on the request, e.g. whether a user is logged in.

tests/ImageSharp.Web.Tests/Helpers/HMACUtilitiesTests.cs
```diff
@@ -10,39 +10,39 @@ public class HMACUtilitiesTests
...
 
         [Theory]
-        [InlineData("The quick brown fox jumped.")]
-        [InlineData("Over the lazy dog.")]
-        public void CanGenerate256BitHMAC(string value)
+        [InlineData("The quick brown fox jumped.", "8ac58e6e983b9cdea7771a2f68b7430c8923ce43f690e23a51192f427fd7ece9")]
+        [InlineData("Over the lazy dog.", "2ea9f37afac5ae93e27d835cc6061ad84b31064ae29f451c9136e88ad5686a1e")]
+        public void CanGenerate256BitHMAC(string value, string expected)
         {
-            string expected = HMACUtilities.ComputeHMACSHA256(value, HMACSecretKey);
```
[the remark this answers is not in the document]
ronaldbarendse: Because checking `Assert.Equal(expected, actual);` doesn't make sense if they're both generated exactly the same way...

The only valuable assertion is whether the length is correct (okay and that it doesn't throw exceptions), but if it would generate a static string of the correct length, this test will still succeed!

PR #265, opened 2022-05-18
Use full URl as HMAC cache key

JimBobSquarePants:
### Prerequisites

### Description

As previously [identified](https://github.com/SixLabors/ImageSharp.Web/pull/250/files#r857755086) using only the token as a cache key opens the potential for abuse of the caching mechanism. PR #256 was opened to remove that cache. 

I would consider removing the cache with current target frameworks to be too expensive given the large overhead of  HMAC generation so have replaced that PR with an implementation that uses the full encoded URL as the key. 

The cache key generation method causes low allocations due to the underlying usage of [`string.Create`](https://github.com/dotnet/aspnetcore/issues/28906) so I would consider it acceptable for now. For .NET 6 we can likely remove the cache in it's entirety due to the much improved reusable hashing methods available to that framework.

### DIFF
--- a/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
+++ b/src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
@@ -246,7 +246,7 @@ namespace SixLabors.ImageSharp.Web.Middleware
 
             ImageCommandContext imageCommandContext = new(httpContext, commands, this.commandParser, this.parserCulture);
 
-            // At this point we know that this is an image request so should attempt to compute a validating HMAC..
+            // At this point we know that this is an image request so should attempt to compute a validating HMAC.
             string hmac = null;
             if (checkHMAC && token != null)
             {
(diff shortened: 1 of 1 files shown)


tocsoft (approved):

ronaldbarendse:
One major downside compared to PR https://github.com/SixLabors/ImageSharp.Web/pull/256 (removing the HMAC cache), is that you can't skip the token validation based on request specific conditions (e.g. by returning `null` from `OnComputeHMACAsync` to indicate no check is required).

You could then use this to skip validation when a user is logged in:
```c#
options.OnComputeHMACAsync = (context, secret) =>
{
    if (context.Context.User.Identity.IsAuthenticated)
    {
        return Task.FromResult<string>(null);
    }

    string uri = CaseHandlingUriBuilder.BuildRelative(
            CaseHandlingUriBuilder.CaseHandling.LowerInvariant,
            context.Context.Request.PathBase,
            context.Context.Request.Path,
            QueryString.Create(context.Commands));

    return Task.FromResult(HMACUtilities.ComputeHMACSHA256(uri, secret));
};
```

JimBobSquarePants:
HMAC validation for requests should always be separate from general user authorization. They really do two different things.

ronaldbarendse:
What about providing a way to skip validation for existing commands/values that you've used? When you currently configure the HMAC secret, all existing image URLs will return 400 error codes...

And will this also work when you've added/removed a command in `OnParseCommandsAsync`? That command won't be in the URL that's used as cache key, but will be used to compute the HMAC token.

JimBobSquarePants:
>What about providing a way to skip validation for existing commands/values that you've used? When you currently configure the HMAC secret, all existing image URLs will return 400 error codes...

Which is absolutely the right thing to do. The HMAC system is all or nothing by design and always should be.. 

>And will this also work when you've added/removed a command in `OnParseCommandsAsync`?

Yes. The token is calculated and compared before `OnParseCommandsAsync`.

PR #266, opened 2022-05-18
Enumerate unordered keys during removal.

JimBobSquarePants:
### Prerequisites

### Description

All credit goes to @ronaldbarendse for this as it was his idea and I've basically copied his implementation. 

Enumerating the underlying unordered key collection removes the overhead of allocating a list for each command collection during a request. Benchmarks show a 4x perf improvement plus 5x reduction in memory.


JimBobSquarePants:
No worries, let's just ditch this then. 

Maybe we can work something out using pooling to avoid the list allocation instead.

PR #268, opened 2022-05-26
Optimize Invalid command removal.

JimBobSquarePants:
### Prerequisites

### Description

This PR does a few things
- Prevents multiple enumeration of comand keys and removes invalid commands in a single loop.
- Allows duplicate commands in the URL (_avoiding ArgumentException_), taking and parsing only the last one. (_Duplicates are not counted in cache key generation as that happens following stripping_)
- Update ImageSharp to v2.1.2

### DIFF
--- a/src/ImageSharp.Web/Commands/PresetOnlyQueryCollectionRequestParser.cs
+++ b/src/ImageSharp.Web/Commands/PresetOnlyQueryCollectionRequestParser.cs
@@ -39,7 +39,8 @@ namespace SixLabors.ImageSharp.Web.Commands
                 return new();
             }
 
-            string requestedPreset = context.Request.Query[QueryKey][0];
+            StringValues query = context.Request.Query[QueryKey];
+            string requestedPreset = query[query.Count - 1];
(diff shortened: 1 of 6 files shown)


brianpopow (approved):
Looks good

ronaldbarendse:
> Allows duplicate commands in the URL (_avoiding ArgumentException_), taking and parsing only the last one. (_Duplicates are not counted in cache key generation as that happens following stripping_)

Can you elaborate how you'd get an `ArgumentException`, because the following query strings both work and will result in the same parsed commands:
- `?width=400&height=400&rxy=1,1`
- `?width=400&height=400&rxy=1&rxy=1`

This is because [`StringValues.ToString()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.stringvalues.tostring?view=dotnet-plat-ext-6.0#microsoft-extensions-primitives-stringvalues-tostring) joins multiple values as a comma separated string, resulting in the same `1,1` value for the `rxy` command, not duplicate commands. This is therefore a breaking change in how the command parsing works, as some users might depend on this behavior.

PR #279, opened 2022-07-01
Url decode path before fetching from file system, fixes #278

### LINKED ISSUE #278: 2.0.0 Breaks on filenames with special characters
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have verified that I am running the latest version of ImageSharp.Web
- [X] I have verified if the problem exist in both `DEBUG` and `RELEASE` mode
- [X] I have searched [open](https://github.com/SixLabors/ImageSharp.Web/issues) and [closed](https://github.com/SixLabors/ImageSharp.Web/issues?q=is%3Aissue+is%3Aclosed) issues to ensure it has not already been reported

### ImageSharp.Web version

2.0.0+

### Other Six Labors packages and versions

n/a

### Environment (Operating system, version and so on)

Linux Mint 20.3

### .NET Framework version

6.0.301

### Description

In 1.0.5, this url: `http://localhost:6123/aśdẁ.jpeg?width=100` resizes the image specified as expected. In 2.0.0 and 2.0.1 the same url doesn't resize the image and yields a debug message `The image 'http://localhost:6123/aśdẁ.jpeg?width=100' could not be resolved`. This is true for other non-ascii characters such as "åäö".

The options for legacy behavior detailed here (https://docs.sixlabors.com/articles/imagesharp.web/gettingstarted.html) with the `LegacyV1CacheKey` doesn't fix it.

### Steps to Reproduce

1. Create a new .net6 asp.net core web application and install ImageSharp.Web 2.0.0 (or 2.0.1). 
2. Add `builder.Services.AddImageSharp()` and `app.UseImageSharp();` to the Program.cs file. 
3. Place an image named aśdẁ.jpg in the wwwroot folder and try accessing it from the browser with a ?width=600 query parameter.

### Images

_No response_

jrunestone:
### Prerequisites

### Description
The base class `FileProviderImageProvider` checks if the unmodified request path exists in the file provider, however this returns false if the path is url encoded (as in the case of special characters in the path, eg á or ö). Simply url decoding the path before checking seems to solve this issue.

Seems reasonable enough? Maybe a test case would be in order.


src/ImageSharp.Web/Providers/FileProviderImageProvider.cs
```diff
@@ -54,7 +55,7 @@ public virtual bool IsValidRequest(HttpContext context)
         /// <inheritdoc/>
         public Task<IImageResolver> GetAsync(HttpContext context)
         {
-            IFileInfo fileInfo = this.fileProvider.GetFileInfo(context.Request.Path);
+            IFileInfo fileInfo = this.fileProvider.GetFileInfo(WebUtility.UrlDecode(context.Request.Path));
```
ronaldbarendse, line 58: It's probably enough to get the underlying value from the `PathString`, as it's actually the implicit `ToString()` that causes it to encode the path:
```suggestion
            IFileInfo fileInfo = this.fileProvider.GetFileInfo(context.Request.Path.Value);
```

See also: https://stackoverflow.com/questions/66471763/inconsistent-url-decoding-of-httprequest-path-in-asp-net-core

jrunestone: OK sure, the potential issue with space/+ mentioned in your link doesn't seem to be a problem (for me) either.

JimBobSquarePants:
@jrunestone I've jut pushed a change to your PR that changes the name of the test image to use diacritics. 

I'll need to make some changes to the tests to ensure the generated token is now correct so please don't make any changes while I work on this.

JimBobSquarePants (approved):
LGTM. Thanks! 👍

jrunestone:
YW, can we get a new release soon? :)

JimBobSquarePants:
There's a new release on NuGet now 😄

PR #281, opened 2022-07-07
Add Authorization Utilitites

JimBobSquarePants:
### Prerequisites

### Description

Adds a new type `ImageSharpRequestAuthorizationUtilities` which completes the HMAC story, allowing developers to create tokens directly using the functions defined in the middleware options.

### DIFF
--- a/src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
+++ b/src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
@@ -126,7 +126,7 @@ namespace SixLabors.ImageSharp.Web.Caching.AWS
         /// <summary>
         /// <see href="https://github.com/aspnet/AspNetIdentity/blob/b7826741279450c58b230ece98bd04b4815beabf/src/Microsoft.AspNet.Identity.Core/AsyncHelper.cs"/>
         /// </summary>
-        internal static class AsyncHelper
+        private static class AsyncHelper
         {
             private static readonly TaskFactory TaskFactory
                 = new(
--- a/src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
+++ b/src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
@@ -1,8 +1,11 @@
 // Copyright (c) Six Labors.
 // Licensed under the Apache License, Version 2.0.
 
+using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
+using System.Threading;
 using System.Threading.Tasks;
 using Amazon.S3;
 using Amazon.S3.Model;
@@ -118,6 +121,54 @@ namespace SixLabors.ImageSharp.Web.Caching.AWS
             }
 
             return null;
+        }
+
+        /// <summary>
+        /// <see href="https://github.com/aspnet/AspNetIdentity/blob/b7826741279450c58b230ece98bd04b4815beabf/src/Microsoft.AspNet.Identity.Core/AsyncHelper.cs"/>
+        /// </summary>
+        internal static class AsyncHelper
+        {
+            private static readonly TaskFactory TaskFactory
+                = new(
+                    CancellationToken.None,
+                    TaskCreationOptions.None,
+                    TaskContinuationOptions.None,
+                    TaskScheduler.Default);
+
+            /// <summary>
+            /// Executes an async <see cref="Task"/> method synchronously.
+            /// </summary>
+            /// <param name="task">The task to excecute.</param>
+            public static void RunSync(Func<Task> task)
+            {
+                CultureInfo cultureUi = CultureInfo.CurrentUICulture;
+                CultureInfo culture = CultureInfo.CurrentCulture;
+                TaskFactory.StartNew(() =>
+                {
+                    Thread.CurrentThread.CurrentCulture = culture;
+                    Thread.CurrentThread.CurrentUICulture = cultureUi;
+                    return task();
+                }).Unwrap().GetAwaiter().GetResult();
+            }
+
+            /// <summary>
+            /// Executes an async <see cref="Task{TResult}"/> method which has
+            /// a <paramref name="task"/> return type synchronously.
+            /// </summary>
+            /// <typeparam name="TResult">The type of result to return.</typeparam>
+            /// <param name="task">The task to excecute.</param>
+            /// <returns>The <typeparamref name="TResult"/>.</returns>
+            public static TResult RunSync<TResult>(Func<Task<TResult>> task)
+            {
+                CultureInfo cultureUi = CultureInfo.CurrentUICulture;
+                CultureInfo culture = CultureInfo.CurrentCulture;
+                return TaskFactory.StartNew(() =>
+                {
+                    Thread.CurrentThread.CurrentCulture = culture;
+                    Thread.CurrentThread.CurrentUICulture = cultureUi;
+                    return task();
+                }).Unwrap().GetAwaiter().GetResult();
+            }
         }
     }
 }
--- a/src/ImageSharp.Web/CaseHandlingUriBuilder.cs
+++ b/src/ImageSharp.Web/CaseHandlingUriBuilder.cs
@@ -14,6 +14,7 @@ namespace SixLabors.ImageSharp.Web
     /// </summary>
     public static class CaseHandlingUriBuilder
     {
+        private static readonly Uri FallbackBaseUri = new("http://localhost/");
         private static readonly SpanAction<char, (bool LowerInvariant, string Host, string PathBase, string Path, string Query)> InitializeAbsoluteUriStringSpanAction = new(InitializeAbsoluteUriString);
 
         /// <summary>
@@ -127,13 +128,11 @@ namespace SixLabors.ImageSharp.Web
             }
             else
             {
-                string components = uri.GetComponents(UriComponents.SerializationInfoString, UriFormat.UriEscaped);
-                if (handling == CaseHandling.LowerInvariant)
-                {
-                    return components.ToLowerInvariant();
-                }
-
-                return components;
+                Uri faux = new(FallbackBaseUri, uri);
+                return BuildRelative(
+                    handling,
+                    pathBase: PathString.FromUriComponent(faux),
+                    query: QueryString.FromUriComponent(faux));
             }
         }
 
--- a/src/ImageSharp.Web/CaseHandlingUriBuilder.cs
+++ b/src/ImageSharp.Web/CaseHandlingUriBuilder.cs
@@ -159,7 +159,7 @@ namespace SixLabors.ImageSharp.Web
                 Uri faux = new(FallbackBaseUri, uri);
                 return BuildRelative(
                     handling,
-                    pathBase: PathString.FromUriComponent(faux),
+                    path: PathString.FromUriComponent(faux),
                     query: QueryString.FromUriComponent(faux));
             }
         }
@@ -210,7 +210,7 @@ namespace SixLabors.ImageSharp.Web
 
             if (uriParts.Scheme.Length > 0)
             {
-                index = CopyTextToBuffer(buffer, index, uriParts.Scheme.AsSpan());
+                index = CopyTextToBufferLowerInvariant(buffer, index, uriParts.Scheme.AsSpan());
                 index = CopyTextToBuffer(buffer, index, Uri.SchemeDelimiter.AsSpan());
             }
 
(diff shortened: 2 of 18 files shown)


src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
```diff
@@ -119,5 +122,53 @@ public Task SetAsync(string key, Stream stream, ImageCacheMetadata metadata)
 
             return null;
         }
+
+        /// <summary>
+        /// <see href="https://github.com/aspnet/AspNetIdentity/blob/b7826741279450c58b230ece98bd04b4815beabf/src/Microsoft.AspNet.Identity.Core/AsyncHelper.cs"/>
+        /// </summary>
+        internal static class AsyncHelper
```
ronaldbarendse, line 129: This should probably be made into a private class or otherwise extracted into its own file:
```suggestion
        private static class AsyncHelper
```

src/ImageSharp.Web/CaseHandlingUriBuilder.cs
```diff
@@ -121,19 +149,18 @@ public static string Encode(CaseHandling handling, Uri uri)
...
-                if (handling == CaseHandling.LowerInvariant)
-                {
-                    return components.ToLowerInvariant();
-                }
-
-                return components;
+                Uri faux = new(FallbackBaseUri, uri);
+                return BuildRelative(
+                    handling,
+                    pathBase: PathString.FromUriComponent(faux),
```
ronaldbarendse, line 162: Make more sense to set the `path`, as the `pathBase` is used as a prefix in ASP.NET Core (and by default also `string.Empty`:
```suggestion
                    path: PathString.FromUriComponent(faux),
```

src/ImageSharp.Web/CaseHandlingUriBuilder.cs
```diff
@@ -181,6 +208,12 @@ private static void InitializeAbsoluteUriString(Span<char> buffer, (bool Lower,
                 pathBaseSpan = pathBaseSpan.Slice(0, pathBaseSpan.Length - 1);
             }
 
+            if (uriParts.Scheme.Length > 0)
+            {
+                index = CopyTextToBuffer(buffer, index, uriParts.Scheme.AsSpan());
```
ronaldbarendse, line 213: The scheme should always be case-insensitive:
```suggestion
                index = CopyTextToBufferLowerInvariant(buffer, index, uriParts.Scheme.AsSpan());
```

src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
```diff
@@ -16,17 +15,15 @@ public sealed class QueryCollectionRequestParser : IRequestParser
...
                 // We return new to ensure the collection is still mutable via events.
                 return new();
             }
 
-            // TODO: Investigate skipping the double allocation here.
-            // In .NET 6 we can directly use the QueryStringEnumerable type and enumerate stright to our command collection
-            Dictionary<string, StringValues> parsed = QueryHelpers.ParseQuery(context.Request.QueryString.ToUriComponent());
             CommandCollection transformed = new();
-            foreach (KeyValuePair<string, StringValues> pair in parsed)
+            foreach (KeyValuePair<string, StringValues> pair in context.Request.Query)
```
ronaldbarendse, line 26: Can't this also use the captured `query` variable?
```suggestion
            foreach (KeyValuePair<string, StringValues> pair in query)
```

src/ImageSharp.Web/ImageSharpRequestAuthorizationUtilities.cs
```diff
@@ -0,0 +1,288 @@
...
+            {
+                return null;
+            }
+
+            CommandCollection commands = this.requestParser.ParseRequestCommands(context.Context);
+            if (handling == CommandHandling.Sanitize)
+            {
+                this.StripUnknownCommands(commands);
+            }
+
```
ronaldbarendse, line 250: This can be removed, as these commands aren't used (they are already parsed/stripped and available on `context.Commands`:
```suggestion
```

ronaldbarendse (changes requested):
This will be a great addition and will make (correct) use of the HMAC image processing protection a lot easier! I've added some comments/suggestions and would love to test this when it's ready for review 👍🏻

src/ImageSharp.Web/CaseHandlingUriBuilder.cs
JimBobSquarePants: Yeah good point.

src/ImageSharp.Web/CaseHandlingUriBuilder.cs
JimBobSquarePants: Yeah, let's do that. 👍

src/ImageSharp.Web/Commands/QueryCollectionRequestParser.cs
JimBobSquarePants: Yeah just missed it.

JimBobSquarePants:
@ronaldbarendse Good feedback, thanks! I've pushed all the fixes.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -144,7 +145,8 @@ public class ImageSharpMiddleware
             ICacheHash cacheHash,
             CommandParser commandParser,
             FormatUtilities formatUtilities,
-            AsyncKeyReaderWriterLock<string> asyncKeyLock)
+            AsyncKeyReaderWriterLock<string> asyncKeyLock,
+            ImageSharpRequestAuthorizationUtilities requestAuthorizationUtilities)
```
ronaldbarendse, line 149: Adding this parameter is a (binary) breaking change which can be avoided by re-adding the previous constructor as obsolete and manually constructing the `ImageSharpRequestAuthorizationUtilities`:

```c#
[Obsolete]
public ImageSharpMiddleware(
    RequestDelegate next,
    IOptions<ImageSharpMiddlewareOptions> options,
    ILoggerFactory loggerFactory,
    IRequestParser requestParser,
    IEnumerable<IImageProvider> resolvers,
    IEnumerable<IImageWebProcessor> processors,
    IImageCache cache,
    ICacheKey cacheKey,
    ICacheHash cacheHash,
    CommandParser commandParser,
    FormatUtilities formatUtilities,
    AsyncKeyReaderWriterLock<string> asyncKeyLock)
    : this(
        next,
        options,
        loggerFactory,
        requestParser,
        resolvers,
        processors,
        cache,
        cacheKey,
        cacheHash,
        commandParser,
        formatUtilities,
        asyncKeyLock,
        new ImageSharpRequestAuthorizationUtilities(options, requestParser, processors, commandParser, null))
{
}
```

This does require `ImageSharpRequestAuthorizationUtilities` to allow a null value for `IServiceProvider`, but that shouldn't be an issue, as it isn't required/used by the methods this middleware calls into anyway.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
```diff
@@ -172,20 +175,10 @@ public class ImageSharpMiddleware
...
-                    commands.Add(command);
-                }
-            }
-
-            this.knownCommands = commands;
-
             this.logger = loggerFactory.CreateLogger<ImageSharpMiddleware>();
             this.formatUtilities = formatUtilities;
             this.asyncKeyLock = asyncKeyLock;
+            this.authorizationUtilities = requestAuthorizationUtilities;
```
ronaldbarendse, line 181: Shouldn't the field also be named `requestAuthorizationUtilities`?

src/ImageSharp.Web/ImageSharpRequestAuthorizationUtilities.cs
```diff
@@ -0,0 +1,277 @@
...
+using SixLabors.ImageSharp.Web.Commands;
+using SixLabors.ImageSharp.Web.Middleware;
+using SixLabors.ImageSharp.Web.Processors;
+
+namespace SixLabors.ImageSharp.Web
+{
+    /// <summary>
+    /// Contains various helper methods for authorizing image requests.
+    /// </summary>
+    public sealed class ImageSharpRequestAuthorizationUtilities
```
ronaldbarendse, line 25: Shouldn't this be renamed to `RequestAuthorizationUtilities`, similar to `FormatUtilities`, `HMACUtilities`, etc?

ronaldbarendse (changes requested):
I've successfully tested this PR together with a massively simplified version of https://github.com/umbraco/Umbraco-CMS/pull/12376 (replacing the custom `ImageSharpImageUrlTokenGenerator` abstraction/implementation with `ImageSharpRequestAuthorizationUtilities`) 🙌🏻

I've only added some final comments about backwards compatibility and naming, but looks good to me otherwise!

src/ImageSharp.Web/ImageSharpRequestAuthorizationUtilities.cs
JimBobSquarePants: I wanted to make it more explicit. It’s for ImageSharp requests only.

src/ImageSharp.Web/Middleware/ImageSharpMiddleware.cs
ronaldbarendse: While looking into this a bit more, I noticed `ImageSharpRequestAuthorizationUtilities` is currently only really required to strip unknown commands, as `ComputeHMACAsync(ImageCommandContext)` just delegates to `this.options.OnComputeHMACAsync(ImageCommandContext, byte[])`. I also think that stripping unknown commands isn't a request authorization concern, so it's probably best to remove this dependency on `ImageSharpRequestAuthorizationUtilities` altogether...

ronaldbarendse:
I've done some further testing, investigation and refactoring to:
- Rename `ImageSharpRequestAuthorizationUtilities` to `RequestAuthorizationUtilities`;
- Remove the dependency on `RequestAuthorizationUtilities` in the middleware (as mentioned in [this comment](https://github.com/SixLabors/ImageSharp.Web/pull/281#discussion_r918390842));
- Obsolete `HMACUtilities.TokenCommand` (as this class isn't responsible for query string/command manipulation and you can now use `RequestAuthorizationUtilities.TokenCommand` instead);
- Include minor fixes like using `CaseHandlingUriBuilder.Encode()` to generate the HMAC token cache key (that ensures the query string/commands are always lowercased), update code styling and do some re-ordering.

You can see all my changes here: https://github.com/SixLabors/ImageSharp.Web/compare/js/auth-utils...ronaldbarendse:ImageSharp.Web:js/auth-utils. I can convert this into a PR targeting `js/auth-utils` if you want to discuss (and easily merge) these, just let me know! 👍🏻

src/ImageSharp.Web/ImageSharpRequestAuthorizationUtilities.cs
ronaldbarendse: Yeah, but it's already in the `SixLabors.ImageSharp.Web` namespace and consumers can otherwise alias the class name if it would conflict with something else:
```c#
using ImageSharpRequestAuthorizationUtilities = SixLabors.ImageSharp.Web.RequestAuthorizationUtilities;
```

And after mentioning that, I acknowledge that you can also use that to do the reverse 😓 So it's more to align with the naming used by all the other utility classes...

JimBobSquarePants:
Thanks for having a crack but I think you missed the point of this new type when making some changes. 

We absolutely want to be using the new type from the middleware. It ensures the same type process is used on both sides. 
We also want to use that type to strip the unknown commands. Whether to expose that method is up for debate. 

Moving operations to various types (`IEnumerable<IImageWebProcessor>`) is never wise, internal or otherwise. 

Tests should be cleaned up, most definitely but the other random changes (logging, field field order) are unnecessary. Code shouldn't be changed for changes sake. We also never, ever use `ObsoleteAttribute` in any of the Six Labors libraries. It's a sure fire way to accumulate technical debt that is unmaintainable given the lack of funding and time.

My changes are only binary breaking if someone is directly calling the middleware constructor which they simply should not be doing. I'll be versioning this release 2.1.0 anyway to send a signal that more fundamental changes have taken place.

src/ImageSharp.Web/ImageSharpRequestAuthorizationUtilities.cs
JimBobSquarePants: The naming is about about consumption intent. All the other types are designed to be used conceptionally within other ImageSharp.Web type implementations (processors generally) whereas this type is designed to be consumed from the users code only. It has to be explicit and unfortunately verbose.

JimBobSquarePants:
This is probably good to merge now.

ronaldbarendse:
@JimBobSquarePants Can this also be released in a 2.0.3 patch version? I'd love to finish the HMAC authorization support in Umbraco (see PR https://github.com/umbraco/Umbraco-CMS/pull/12376) 😇

ronaldbarendse:
I see the work on the `TagHelper` PR has stalled a bit and has currently has some unresolved comments, but I would really like to make use of the new authorization utilities. Would it be possible for you to create a 2.0.3 release containing just these changes? 🙏🏻

JimBobSquarePants:
Let me see what Ican do. No promises though. I've got a lot on just now.

PR #282, opened 2022-07-08
Added a check when forbidden is returned from S3 bucket

mdupras:
…sts in a S3 Bucket.

### Prerequisites

### Description
We don't need the `ListBucket` permission when accessing the S3 bucket. We will always know the key we want to access, but the side effect is we are getting a Forbidden from the S3 bucket, probably to prevent bucket enumeration. I think it's safe to assume that the Key doesn't exist.

### DIFF
--- a/src/ImageSharp.Web.Providers.AWS/Providers/AWSS3StorageImageProvider.cs
+++ b/src/ImageSharp.Web.Providers.AWS/Providers/AWSS3StorageImageProvider.cs
@@ -161,6 +161,12 @@ namespace SixLabors.ImageSharp.Web.Providers.AWS
                     return false;
                 }
 
+                // If the object exists but the client is not authorized to access it, then a "Forbidden" will be thrown.
+                if (string.Equals(e.ErrorCode, "Forbidden"))
+                {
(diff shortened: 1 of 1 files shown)


JimBobSquarePants:
Thanks for this. Is there any way we can test it?

JimBobSquarePants:
Ah, I meant in our unit tests. If not I'll just trust you. I have no way to test and little experience with S3.

mdupras:
Ahh I'm sorry, I didn't think it was possible with the current setup.

I have just validated it and I'm getting this error : `A parameter you provided implies functionality that is not implemented` when I try to set a policy on the bucket. At the very least, s3rver would need to have some permissions support and I don't think it has any right now.

If you want to test the PR, it would be outside the CI, with a real bucket. I can make you some terraform files for a repro env if you want, but you would need a working **AWS Account**.

JimBobSquarePants:
No worries. Looks like you've covered everything. Unfortunately the test strategy for AWS isn't fully comprehensive.

JimBobSquarePants (approved):

PR #287, opened 2022-09-26
Aws s3 client timeout

mdupras:
### Prerequisites

### Description

Most configurations of the S3Client can be set via environment variable, but it is not the case for the Timeout settings. I've only added a config and forwarded it.


JimBobSquarePants (approved):
Looks great! Thanks for your help here!

PR #290, opened 2022-10-05
Introduce Taghelper

JimBobSquarePants:
### Prerequisites

### Description

Adds a new `ImageSharpTaghelper` that greatly simplifies URL generation (_including automatic generation of HMAC keys_). 
The TagHelper is designed to be inheritable allowing the addition of custom commands.

Example usage (_note, the syntax highlighter here does not do the TagHelper justice. The property values are all strong typed_)

```razor
<img src="sixlabors.imagesharp.web.png"
     width="300"
     height="200"
     imagesharp-width="600"
     imagesharp-height="400"
     imagesharp-rmode="ResizeMode.Pad"
     imagesharp-rcolor="Color.LimeGreen" />
```

The TagHelper will only intitialize when both the following criteria are met: 
1. The TagHelper is explicitly referenced in _ViewImports.cshtml via `@addTagHelper *, SixLabors.ImageSharp.Web`
2. The `src` and ` imagesharp-` prefixed HTML attributes are declared.


src/ImageSharp.Web/ImageSharpTagHelper.cs
```diff
@@ -0,0 +1,363 @@
...
+
+        private void AddQualityCommands(CommandCollection commands)
+        {
+            if (this.Quality.HasValue)
+            {
+                commands.Add(QualityWebProcessor.Quality, this.Quality.Value.ToString(this.parserCulture));
+            }
+        }
+
+        private static string AddQueryString(
```
ronaldbarendse, line 324: Can't you use [`QueryHelpers.AddQueryString()`](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.webutilities.queryhelpers.addquerystring?view=aspnetcore-6.0) instead?

ronaldbarendse (changes requested):
This is a really cool feature and will make it a _LOT_ easier to generate the correct URLs 🙌🏻

I've added some comments/suggestions and just having support for `` is already great for the initial release. Adding `srcset` support for both `img` and `picture/source` elements  will probably be out of scope for this PR, since there's a lot of ways you can render [responsive images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images).

src/ImageSharp.Web/Format.cs
```diff
@@ -0,0 +1,41 @@
+// Copyright (c) Six Labors.
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Web
+{
+    /// <summary>
+    /// Provides enumeration of possible image formats.
+    /// </summary>
+    public enum Format
```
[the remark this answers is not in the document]
JimBobSquarePants: Yep exactly. I could tackle it another way though and use structs like we do with resampler implementations in the core library. I’ll have a crack.

src/ImageSharp.Web/ImageSharpTagHelper.cs
```diff
@@ -0,0 +1,363 @@
...
+    [HtmlTargetElement("img", Attributes = SrcAttributeName + "," + OrientAttributeName, TagStructure = TagStructure.WithoutEndTag)]
+    [HtmlTargetElement("img", Attributes = SrcAttributeName + "," + AutoOrientAttributeName, TagStructure = TagStructure.WithoutEndTag)]
+    [HtmlTargetElement("img", Attributes = SrcAttributeName + "," + FormatAttributeName, TagStructure = TagStructure.WithoutEndTag)]
+    [HtmlTargetElement("img", Attributes = SrcAttributeName + "," + BgColorAttributeName, TagStructure = TagStructure.WithoutEndTag)]
+    [HtmlTargetElement("img", Attributes = SrcAttributeName + "," + QualityAttributeName, TagStructure = TagStructure.WithoutEndTag)]
+    public class ImageSharpTagHelper : UrlResolutionTagHelper
+    {
+        private const string SrcAttributeName = "src";
+        private const string WidthAttributeName = ResizeWebProcessor.Width;
+        private const string HeightAttributeName = ResizeWebProcessor.Height;
```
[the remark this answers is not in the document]
JimBobSquarePants: That was actually a conscious decision. My thought process was that it’s bad practice to use width/height attributes with unscaled content. Happy to be proven that this is a bad idea though.

src/ImageSharp.Web/ImageSharpTagHelper.cs
JimBobSquarePants: Ah no, cos that expects a name value collection. This is copied straight from that anyway.

src/ImageSharp.Web/Format.cs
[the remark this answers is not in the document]
JimBobSquarePants: Now uses a struct with static defaults in a separate container class.

samples/ImageSharp.Web.Sample/Pages/_ViewImports.cshtml
```diff
@@ -0,0 +1,7 @@
+@using ImageSharp.Web.Sample
```
JimBobSquarePants, line 1: It would be nice if we could add these namespaces at startup however, I don't think it's possible.

src/ImageSharp.Web/ImageSharpTagHelper.cs
```diff
@@ -0,0 +1,363 @@
...
+        private const string AttributePrefix = "asp-";
```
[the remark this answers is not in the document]
JimBobSquarePants: `imagesharp-` is the new prefix. Verbose it better for discovery and intellisense helps with generation.

src/ImageSharp.Web/ImageSharpTagHelper.cs
```diff
@@ -0,0 +1,365 @@
...
+        {
+            int anchorIndex = uri.IndexOf('#');
+            string uriToBeAppended = uri;
+            string anchorText = string.Empty;
+
+            // If there is an anchor, then the query string must be inserted before its first occurrence.
+            if (anchorIndex != -1)
+            {
+                anchorText = uri.Substring(anchorIndex);
+                uriToBeAppended = uri.Substring(0, anchorIndex);
```
deanmarcussen, line 338: span? could look at QueryHelpers.AddQueryString for the span variant

JimBobSquarePants: Good catch. Much better now!

src/ImageSharp.Web/Format.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        public static FormatCommand Gif { get; } = new(nameof(Gif));
+
+        /// <summary>
+        /// Gets the Jpg encoding format.
+        /// </summary>
+        public static FormatCommand Jpg { get; } = new(nameof(Jpg));
+
+        /// <summary>
+        /// Gets the Gif encoding format.
```
brianpopow, line 27:
```suggestion
        /// Gets the Png encoding format.
```

src/ImageSharp.Web/Format.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        public static FormatCommand Png { get; } = new(nameof(Png));
+
+        /// <summary>
+        /// Gets the Bmp encoding format.
```
brianpopow, line 32:
```suggestion
        /// Gets the Tga encoding format.
```

src/ImageSharp.Web/Format.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// </summary>
+        public static FormatCommand Tga { get; } = new(nameof(Tga));
+
+        /// <summary>
+        /// Gets the Gif encoding format.
```
brianpopow, line 37:
```suggestion
        /// Gets the WebP encoding format.
```

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
ronaldbarendse: My main issue is that this will result in changing all existing image URLs to include the width/height (and HMAC token) commands by simply adding `@addTagHelper *, SixLabors.ImageSharp.Web` to your `_ViewImports.cshtml`. It will also process and cache these images, even though they will most likely already be the correct size (otherwise they would get scaled or distorted, which might even be done on purpose, e.g. when providing a high DPI version, upscaling, etc.).

Besides that, you might want to use a custom command to do the resizing (like Umbraco does with crop coordinates in the `CropWebProcessor`) and omit the width command from the URL while still explicitly setting the width attribute on the `img` tag. Another case might be when you disable the auto-orient feature and therefore have to swap the width/height commands.

You can avoid this by also using prefixes for these commands, so `` will be rendered as ``. This will still set the attributes, but only resize the image when explicitly set. This will also allow adding high DPI versions by using ``.

samples/ImageSharp.Web.Sample/Pages/_ViewImports.cshtml
ronaldbarendse: You can add [implicit/global usings](https://learn.microsoft.com/en-us/dotnet/core/tutorials/top-level-templates#implicit-using-directives) by providing a build target file that contains:
```xml
<ItemGroup>
  <Using Include="SixLabors.ImageSharp" />
  <Using Include="SixLabors.ImageSharp.Processing" />
  <Using Include="SixLabors.ImageSharp.Web" />
</ItemGroup>
```

Umbraco does this as well: https://github.com/umbraco/Umbraco-CMS/blob/3873a0735cde27bfd9692f4c7370613730ef4fd0/src/Umbraco.Cms.Targets/buildTransitive/Umbraco.Cms.Targets.props#L8-L11.

src/ImageSharp.Web/ImageSharpTagHelper.cs
```diff
@@ -0,0 +1,366 @@
...
+using Microsoft.AspNetCore.Mvc.Routing;
+using Microsoft.AspNetCore.Mvc.TagHelpers;
+using Microsoft.AspNetCore.Razor.TagHelpers;
+using Microsoft.Extensions.Options;
+using SixLabors.ImageSharp.Processing;
+using SixLabors.ImageSharp.Web.Commands;
+using SixLabors.ImageSharp.Web.Middleware;
+using SixLabors.ImageSharp.Web.Processors;
+
+namespace SixLabors.ImageSharp.Web
```
ronaldbarendse, line 19: I think this class should be renamed to `ImageTagHelper` and the namespace changed to:
```suggestion
namespace SixLabors.ImageSharp.Web.TagHelpers
```

It might not be a big deal now, but if you'd like to add additional tag helpers later (e.g. picture/source or srcset), removing `ImageSharp` from the current class will help in naming them and also makes conditionally adding them to your view a little bit easier and non-repetitive (e.g. `@addTagHelper SixLabors.ImageSharp.Web.TagHelpers.ImageTagHelper, SixLabors.ImageSharp.Web`).

src/ImageSharp.Web/ImageSharpTagHelper.cs
```diff
@@ -0,0 +1,366 @@
...
+            {
+                if (parameter.Value is null)
+                {
+                    continue;
+                }
+
+                sb.Append(hasQuery ? '&' : '?');
+                sb.Append(UrlEncoder.Default.Encode(parameter.Key));
+                sb.Append('=');
+                sb.Append(UrlEncoder.Default.Encode(parameter.Value));
```
ronaldbarendse, line 358: I noticed all command keys are lowercase, but the values (especially for enums) are capitalized. Even though the commands are case-insensitive and work the same regardless, I think it's nicer (at least to the eye) to have all lowercase values as well:
```suggestion
                sb.Append(UrlEncoder.Default.Encode(parameter.Value.ToLowerInvariant()));
```

samples/ImageSharp.Web.Sample/Pages/_ViewImports.cshtml
JimBobSquarePants: I'm a daft one. I already know this! 🤣

https://github.com/SixLabors/ImageSharp/blob/451a713fc11f2825946350c8f02c25b310509f57/SixLabors.ImageSharp.props

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
JimBobSquarePants: OK. You've convinced me. Explicit it is.

src/ImageSharp.Web/ImageSharpTagHelper.cs
JimBobSquarePants: It's also 2X more expensive.

src/ImageSharp.Web/ImageSharpTagHelper.cs
JimBobSquarePants: Pushed the suggestion by accident there. Strongly disagree. They write this stuff one time. I'm not reducing the discoverability of type.

JimBobSquarePants: Great suggestion! I've implemented that now.

samples/ImageSharp.Web.Sample/Pages/_ViewImports.cshtml
JimBobSquarePants: Done. Those includes in Umbraco should only be enabled if `ImplicitUsings` is enabled btw.

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
JimBobSquarePants: Just to clarify the new behavior.

- We only resize if `imagesharp-width` or `imagesharp-height` are explicitly set. 
- We only set the width or height attributes if there are no attributes present.

samples/ImageSharp.Web.Sample/Pages/_ViewImports.cshtml
ronaldbarendse: Sometimes you just need a quick reminder! 😉 And thanks for the suggestion, I've just created a PR to fix this in Umbraco: https://github.com/umbraco/Umbraco-CMS/pull/13322 💪🏻

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
ronaldbarendse: Correct, so:
- `` becomes:
  ``
- `` becomes:
  ``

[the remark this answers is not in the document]
ronaldbarendse: The only thing that's not possible is only adding the width or height command without setting the attributes, although I can't think of any use cases where that would make sense...

I did notice the HMAC token doesn't include the query string/commands that are already in the src URL. It's probably a good idea to include that and provide a way to opt-in to adding the HMAC token using e.g `imagesharp-token="true"` or `imagesharp-hmac="true"`, in case no other tag helper attributes are use. This will allow easy HMAC token generation to URLs that already include ImageSharp commands.

samples/ImageSharp.Web.Sample/Program.cs
```diff
@@ -1,29 +1,77 @@
...
-        /// Creates an <see cref="IHostBuilder"/> instance used to create a configured <see cref="IHost"/>.
-        /// </summary>
-        /// <param name="args">Any arguments to pass to the application.</param>
-        /// <returns>The <see cref="IHostBuilder"/>.</returns>
-        public static IHostBuilder CreateHostBuilder(string[] args) =>
-            Host.CreateDefaultBuilder(args)
-                .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>());
+            // Add services to the container.
+            IServiceCollection services = builder.Services;
+            services.AddRazorPages();
```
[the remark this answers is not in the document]
JimBobSquarePants: Yeah, that's my bad. Playing about with a CORS handling solution in another branch and managed to commit code to this one by accident.

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
JimBobSquarePants: We don't need to do the first point since the cache key and HMAC are created only from the commands and not the raw querystring so additional unrecognized commands are ignored.

Regarding the second point. Yeah, we should protect all request coming to providers.

Adding `imagesharp-hmac` is enough now, values passed to the attribute are ignored.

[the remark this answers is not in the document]
ronaldbarendse: Looks like the query string/commands from the original `src` attribute aren't included in the HMAC token generation, as the following returns a 400 status code: ``. This currently makes the addition of the `imagesharp-hmac` attribute rather useless, as using any of the other `imagesharp-` attributes will already ensure the token is added when required.

If the commands are parsed from the `src` attribute, the tag helper could automatically append the HMAC token when a HMAC secret is configured and the URL contains known commands, without having to add any other `imagesharp-` attributes. This would make enabling HMAC processing protection on your existing site really easy, as you'd only have to ensure the tag helper is added to ensure all image URLs in your Razor views will have the correct HMAC tokens added.

This functionality could also be implemented for `picture/source` elements and `srcset` attributes. It might then be a good idea to separate this functionality into a different tag helper, so you can opt-in by adding e.g. `@addTagHelper SixLabors.ImageSharp.Web.TagHelpers.HmacTokenTagHelper, SixLabors.ImageSharp.Web` (and set the [`Order` property](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.razor.taghelpers.taghelper.order) to ensure it's processed after the `ImageTagHelper`). This avoids the URL processing hit in the `ImageTagHelper` and allows you to remove the dependency on `RequestAuthorizationUtilities` from that tag helper as well...

Another benefit would be that adding commands to the URL in the `src` attribute enables you to add ImageSharp commands in front of the ones added by the `imagesharp-` attributes, allowing you to change the order (which is important, as this impacts how the image is processed).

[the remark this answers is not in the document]
JimBobSquarePants: Eh? Why the hell would someone write a command like that mixed in with the helper syntax? They do that and they’re on their own cos it’s bloody stupid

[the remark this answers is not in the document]
JimBobSquarePants: Use the helper, don’t use the helper fine by me but don’t expect me to support stupid behaviour.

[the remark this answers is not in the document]
ronaldbarendse: I may not have explained it clearly enough 😇 There's a couple of benefits when separating the HMAC token generation into a separate tag helper:
1. You don't have to create your own tag helper (inherited from `ImageTagHelper`) to add a custom command (e.g. `imagesharp-cropcoordinates`) and have it included in the HMAC token, as you can then manually include that command in the image URL/`src` attribute;
2. The HMAC tokens can easily be added to other elements and attributes (e.g. `img[srcset]` and `source[srcset]`), without having to implement new tag helpers to generate the image URLs for these elements/attributes;
3. Existing image URLs that contain valid commands (e.g. returned from a CMS) can be augmented with additional commands and/or the HMAC token.

Maybe some code might help to explain it, so please have a look at these changes: https://github.com/SixLabors/ImageSharp.Web/commit/c72ee4a890759b5de4dd0357acaa51b059862dc8 (I've added some other suggestions to my fork, [see branch comparison](https://github.com/SixLabors/ImageSharp.Web/compare/js/taghelper...ronaldbarendse:ImageSharp.Web:js/taghelper)).

src/ImageSharp.Web/TagHelpers/ImageTagHelper.cs
```diff
@@ -0,0 +1,378 @@
...
+                commands.Add(ResizeWebProcessor.Width, this.Width.Value.ToString(this.parserCulture));
+                output.Attributes.SetAttribute(ResizeWebProcessor.Width, width ?? this.Width);
+            }
+
+            int? height = output.Attributes[ResizeWebProcessor.Height]?.Value as int?;
+            if (this.Height.HasValue)
+            {
+                commands.Add(ResizeWebProcessor.Height, this.Height.Value.ToString(this.parserCulture));
+                output.Attributes.SetAttribute(ResizeWebProcessor.Height, height ?? this.Height);
+            }
```
ronaldbarendse, line 268: The attribute value of the width/height will most likely be a `string` (not an `int?`), so it currently still overwrites the attribute values. There's also no reason to get this value if the `imagescharp-` prefixed values aren't set, so this can be rewritten as:

```suggestion
            if (this.Width.HasValue)
            {
                commands.Add(ResizeWebProcessor.Width, this.Width.Value.ToString(this.parserCulture));

                if (!output.Attributes.ContainsName(ResizeWebProcessor.Width))
                {
                    output.Attributes.SetAttribute(ResizeWebProcessor.Width, this.Width.Value);
                }
            }
            
            if (this.Height.HasValue)
            {
                commands.Add(ResizeWebProcessor.Height, this.Height.Value.ToString(this.parserCulture));

                if (!output.Attributes.ContainsName(ResizeWebProcessor.Height))
                {
                    output.Attributes.SetAttribute(ResizeWebProcessor.Height, this.Height.Value);
                }
            }
```

JimBobSquarePants: >The attribute value of the width/height will most likely be a string (not an int?)

No, they're not. The specification described width and height of image elements as representative of CSS pixel units. They're not strings.

src/ImageSharp.Web/Format.cs
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the Apache License, Version 2.0.
+
+namespace SixLabors.ImageSharp.Web
+{
+    /// <summary>
+    /// Contains reusable static instances of known image formats.
+    /// </summary>
+    public static class Format
+    {
+        /// <summary>
```
MarkCiliaVincenti, line 11: What about others like APNG, AVIF and even SVG?

JimBobSquarePants: You can create new `FormatCommand` instances for additional raster types to pass as a parameter. This static class simply holds known values. 

Note: We'd never support SVG since it is not a raster image format.

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
JimBobSquarePants: @ronaldbarendse if you're still keen, I'd like to revisit your suggestion to split out the helper?

src/ImageSharp.Web/TagHelpers/ImageTagHelper.cs
JimBobSquarePants: We always want to read the values anyway and compare as an explicit with might have been set that is different to the requested width. (Think downscaling for high resolution)

src/ImageSharp.Web/ImageSharpTagHelper.cs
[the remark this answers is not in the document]
JimBobSquarePants: OK. I ran with it. 

I have implemented pretty much everything you suggested with a few implementation tweaks. 

HMAC is now the separate helper you designed and requests with no commands no longer require HMAC.

PR #298, opened 2023-01-04
Update .NET

stefannikolei:
Removed everything for netcoreapp3.1 and netcoreapp2.1. Those are out of support.

Added .NET 7.0 as target

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -28,15 +28,15 @@ jobs:
             runtime: -x64
             codecov: true
           - os: ubuntu-latest
-            framework: netcoreapp3.1
+            framework: net7.0
             runtime: -x64
             codecov: false
           - os: macos-latest
-            framework: netcoreapp3.1
+            framework: net7.0
             runtime: -x64
             codecov: false
           - os: windows-latest
-            framework: netcoreapp3.1
+            framework: net7.0
             runtime: -x64
             codecov: false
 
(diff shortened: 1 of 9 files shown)


sebastienros:
There are some misconceptions in this PR, but some things are still valid.

Testing on net7.0 instead of net6.0 is probably fine (though maybe testing on both is even better).
However, the main issue is that for libraries the TFM doesn't represent which version it is compatible with, but the lowest version it requires. The TFM of the application that uses the library will state which final version is used. You only need to target newer or different TFMs in libraries if you have conditional compilation declarations or custom references. 

In your changes you didn't introduce any of these custom dependencies, so the changes are not necessary. It means that by keeping netcoreapp3.1 you can allow anyone to use it for any TFM above to netcoreapp3.1. The only thing you did is prevent people using older TFMs from using it. I know it's not supported anymore, but since there is no technical reason to require a higher one (yet) I don't think updating it is necessary.

NB: This is my opinion and the maintainers might have a different one.

JimBobSquarePants:
@stefannikolei Thanks for the PR. @sebastienros is correct here though as incrementing the target framework is not something we would do normally.

However, I want to move development to target the new ImageSharp v3 development builds. 

What we need to do is create a release/2.0.x branch to represent the latest release and migrate the main branch to target the development builds. This would also require a fair amount of work as an update to the shared submodule and nullability changes would be required also.

JimBobSquarePants:
@stefannikolei I've merged #303 to replace this. We'll be targeting ImageSharp V3 builds in the near future.

PR #306, opened 2023-03-01
Add Parsing of TargetRectangle to ResizeWebProcessor

stefannikolei:
### Prerequisites

### Description
Adds the ability to specify the targetrectangle in the commands. As asked in #302

### DIFF
--- a/src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
+++ b/src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
@@ -62,16 +62,6 @@ public class ResizeWebProcessor : IImageWebProcessor
     /// </summary>
     public const string Compand = "compand";
 
-    /// <summary>
-    /// The command constant for the resize TargetRectangle X
-    /// </summary>
-    public const string TargetRectangleWidth = "targetrectanglewidth";
-
-    /// <summary>
-    /// The command constant for the resize TargetRectangle X
-    /// </summary>
-    public const string TargetRectangleHeight = "targetrectangleheight";
-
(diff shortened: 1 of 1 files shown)


stefannikolei:
Attention this targets the v3 branch

src/ImageSharp.Web/Processors/ResizeWebProcessor.cs
```diff
@@ -134,7 +157,8 @@ public class ResizeWebProcessor : IImageWebProcessor
             Mode = GetMode(commands, parser, culture),
             Compand = GetCompandMode(commands, parser, culture),
             Sampler = GetSampler(commands),
-            PadColor = parser.ParseValue<Color>(commands.GetValueOrDefault(Color), culture)
+            PadColor = parser.ParseValue<Color>(commands.GetValueOrDefault(Color), culture),
+            TargetRectangle = targetRectangle
```
JimBobSquarePants, line 161: I think we can just use `new(0 ,0, size.Width, size.Height)`. We don't do source/target rectangles in this library.

JimBobSquarePants: Ah no... 

I meant don't add these sommnds at all and use the already parsed width/height.

stefannikolei: But then the rectangle will always be created when a width and height is set. Would that be ok. When we add them we are probably more flexible, or doesn't it make sense to use different once?

JimBobSquarePants (approved):
Looks great. Thanks! 👍

PR #309, opened 2023-03-05
Remove nullable disable from azure

stefannikolei:
### Prerequisites

### Description
* Removed all Nullable disable from the azure providers
* Removed IAzureBlobContainerClientOptions because it wasn't used just for declaration
* Added convenience DI methods for Adding the cache or image provider. Leaving out the initialization was on purpose to not run that code on startup because it could throw an exception so the app won't start..


src/ImageSharp.Web.Providers.Azure/SeviceCollectionExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+using SixLabors.ImageSharp.Web.Caching.Azure;
+using SixLabors.ImageSharp.Web.DependencyInjection;
+using SixLabors.ImageSharp.Web.Providers.Azure;
+
+namespace SixLabors.ImageSharp.Web;
+
+/// <summary>
+/// Extension methods for <see cref="IServiceCollection"/> to simplify middleware service registration.
+/// </summary>
+public static class SeviceCollectionExtensions
```
JimBobSquarePants, line 14: I don't think we need these extensions. I'd rather we kept the API consistent across all provider types.

stefannikolei: Ok. Gonna remove that file again. But I think that there is a error on the website. 

image

This must be .SetCache<Adblob..>() or am I missing something?

JimBobSquarePants: The `<` `>` tags were not HTML encoded. Have updated them now thanks.

PR #312, opened 2023-03-06
Remove nullable disable from aws

stefannikolei:
### Prerequisites

- [ ] I have written a descriptive pull-request title
- [ ] I have verified that there are no overlapping [pull-requests](https://github.com/SixLabors/ImageSharp.Web/pulls) open

### DIFF
--- a/src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
+++ b/src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
@@ -1,6 +1,5 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
-#nullable disable
 
 using Amazon;
 using Amazon.Runtime;
@@ -33,7 +32,7 @@ internal static class AmazonS3ClientFactory
         else if (!string.IsNullOrWhiteSpace(options.AccessKey))
         {
             // AccessSecret can be empty.
-            Guard.NotNullOrWhiteSpace(options.Region, nameof(options.Region));
+            Guard.NotNullOrWhiteSpace(options.Region!, nameof(options.Region));
             RegionEndpoint region = RegionEndpoint.GetBySystemName(options.Region);
             AmazonS3Config config = new() { RegionEndpoint = region, UseAccelerateEndpoint = options.UseAccelerateEndpoint };
             SetTimeout(config, options.Timeout);
--- a/src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
+++ b/src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
@@ -32,7 +32,7 @@ internal static class AmazonS3ClientFactory
         else if (!string.IsNullOrWhiteSpace(options.AccessKey))
         {
             // AccessSecret can be empty.
-            Guard.NotNullOrWhiteSpace(options.Region!, nameof(options.Region));
+            Guard.NotNullOrWhiteSpace(options.Region, nameof(options.Region));
             RegionEndpoint region = RegionEndpoint.GetBySystemName(options.Region);
             AmazonS3Config config = new() { RegionEndpoint = region, UseAccelerateEndpoint = options.UseAccelerateEndpoint };
             SetTimeout(config, options.Timeout);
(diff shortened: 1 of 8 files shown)


src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
```diff
@@ -33,7 +32,7 @@ public static AmazonS3Client CreateClient(IAWSS3BucketClientOptions options)
         else if (!string.IsNullOrWhiteSpace(options.AccessKey))
         {
             // AccessSecret can be empty.
-            Guard.NotNullOrWhiteSpace(options.Region, nameof(options.Region));
+            Guard.NotNullOrWhiteSpace(options.Region!, nameof(options.Region));
```
JimBobSquarePants, line 35: This reads like we need to update `Guard`? We cannot assert a string is non-null when passing it as a parameter for a null check.

src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
```diff
@@ -17,7 +16,7 @@ namespace SixLabors.ImageSharp.Web.Caching.AWS;
 public class AWSS3StorageCache : IImageCache
 {
     private readonly IAmazonS3 amazonS3Client;
-    private readonly string bucket;
+    private readonly string? bucket;
```
JimBobSquarePants, line 19: We should do a guard check for the bucket name and make this non-nullable.

src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
```diff
@@ -33,7 +32,7 @@ public static AmazonS3Client CreateClient(IAWSS3BucketClientOptions options)
...
             RegionEndpoint region = RegionEndpoint.GetBySystemName(options.Region);
```
JimBobSquarePants, line 36: The assertion should be on this line by the looks of things.

src/ImageSharp.Web.Providers.AWS/IAWSS3BucketClientOptions.cs
```diff
@@ -24,20 +23,20 @@ internal interface IAWSS3BucketClientOptions
...
     /// <summary>
     /// Gets or sets the AWS secret - Can be used to override keys provided by the environment.
     /// If deploying inside an EC2 instance AWS keys will already be available via environment
     /// variables and don't need to be specified. Follow AWS best security practices on  <see href="https://docs.aws.amazon.com/general/latest/gr/aws-access-keys-best-practices.html"/>.
     /// </summary>
-    string AccessSecret { get; set; }
+    string? AccessSecret { get; set; }
 
     /// <summary>
     /// Gets or sets the AWS endpoint - used for testing to over region endpoint allowing it
```
JimBobSquarePants, line 36: Can we please change the docs comment here to the following.
>Gets or sets the AWS endpoint - used to override the default service endpoint.

src/ImageSharp.Web.Providers.AWS/Providers/AWSS3StorageImageProviderOptions.cs
```diff
@@ -21,19 +20,19 @@ public class AWSS3StorageImageProviderOptions
 public class AWSS3BucketClientOptions : IAWSS3BucketClientOptions
 {
     /// <inheritdoc/>
-    public string Region { get; set; }
+    public string? Region { get; set; }
 
     /// <inheritdoc/>
-    public string BucketName { get; set; }
+    public string BucketName { get; set; } = null!;
```
JimBobSquarePants, line 26: Perhaps we should switch to `get; init;` for our options types to avoid these kinds of hacks. (We're .NET 6)

stefannikolei: That makes no difference. Init does not mean that he must set it. It just says that it can not be changed after constructing. 

For your desired behavior we need the required keyword and that is net7...

JimBobSquarePants: Yeah, you're right (we keep having this conversation 🤣)

I do thoink for .NET 6 though, options-types should use this pattern to enforce immutability.

src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
stefannikolei: we guard for the settings.region and not the created region. So this should be ok

src/ImageSharp.Web.Providers.AWS/Caching/AWSS3StorageCache.cs
stefannikolei: It is non nullable. But to achieve this we have the ugly hack with = null! in the options :(

JimBobSquarePants (approved):
Perfect. Thanks!

PR #313, opened 2023-03-06
Add GrayscaleWebProcessor

stefannikolei:
### Prerequisites

### Description
Adds an Grayscale processor as in #310


JimBobSquarePants:
Thanks, but I don't want to add this. 

If we add everything people ask for then we vastly increase the attack surface for DDOS. The library is extensible and they have all the tools to add their own processors.

PR #317, opened 2023-03-15
Updated the `KeyExists` from aws to keep the response metadata in the image resolver

mdupras:
### Prerequisites

### Description
When the image is resolved we validate if it exists with the method `KeyExists`. To validate this the methods fetch all the object metadata, if it fails it's considered not existent, but the result of that query isn't used.
This PR keeps the result and pass it to the AWS S3 image resolver.
I added an optional parameter for that metadata, in case the resolver needs to be build it elsewhere. If the metadata is not provided, it will fetch it like it used to.

Since I had overlapped change with https://github.com/SixLabors/ImageSharp.Web/pull/312, I included the change of those files in my PR to limit the conflicts and use the nullable options ;)

### DIFF
--- a/src/ImageSharp.Web.Providers.AWS/Providers/AWSS3StorageImageProvider.cs
+++ b/src/ImageSharp.Web.Providers.AWS/Providers/AWSS3StorageImageProvider.cs
@@ -1,6 +1,5 @@
 // Copyright (c) Six Labors.
 // Licensed under the Six Labors Split License.
-#nullable disable
 
 using Amazon.S3;
 using Amazon.S3.Model;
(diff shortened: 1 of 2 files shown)


JimBobSquarePants:
HI @mdupras thanks for this. 

Would it be possible to revert the nullability changes since they overlap with the other PR which will actually increase the probability of conflicts since there have been changes requested there?

mdupras:
> HI @mdupras thanks for this.
> 
> Would it be possible to revert the nullability changes since they overlap with the other PR which will actually increase the probability of conflicts since there have been changes requested there?

Yea no problem! Just did it!

JimBobSquarePants (approved):
Lovely stuff. Thanks!

PR #330, opened 2023-06-23
Use HMAC one-shot methods instead of instances

jevgenigeurtsen:
### Prerequisites

### Description
As recommended by Microsoft (ref: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1850  
 and https://github.com/dotnet/runtime/issues/40579) we should favor the one-shot `HashData` over the `ComputeHash` methods whilst hashing HMAC hashes.

I've modified the source code and ran some benchmarks, the performance gain is minimal, but its still there:
```
|   Method |      Job |  Runtime | sizeKey | sizeInput |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|--------- |--------- |--------- |-------- |---------- |---------:|---------:|---------:|-------:|----------:|
| Instance | .NET 6.0 | .NET 6.0 |      64 |        96 | 581.8 ns | 11.36 ns | 16.29 ns | 0.0420 |     536 B |
|  OneShot | .NET 6.0 | .NET 6.0 |      64 |        96 | 356.7 ns |  7.15 ns |  7.65 ns | 0.0119 |     152 B |
| Instance | .NET 7.0 | .NET 7.0 |      64 |        96 | 604.6 ns | 10.31 ns |  9.64 ns | 0.0420 |     528 B |
|  OneShot | .NET 7.0 | .NET 7.0 |      64 |        96 | 361.9 ns |  6.17 ns |  5.47 ns | 0.0119 |     152 B |
```

Cheers!

### DIFF
--- a/src/ImageSharp.Web/HMACUtilities.cs
+++ b/src/ImageSharp.Web/HMACUtilities.cs
@@ -20,6 +20,16 @@ public static class HMACUtilities
     public const string TokenCommand = "hmac";
 
     /// <summary>
+    /// HMAC hash algorithms and their respective hash sizes in bytes.
+    /// </summary>
+    private enum HashAlgorithmSizes
+    {
+        HMACSHA256 = 32,
+        HMACSHA384 = 48,
+        HMACSHA512 = 64
+    }
+
+    /// <summary>
     /// Computes a Hash-based Message Authentication Code (HMAC) by using the SHA256 hash function.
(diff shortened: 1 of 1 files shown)


JimBobSquarePants (approved):
Ah this is great! The difference here will absolutely add up. Thanks!

PR #341, opened 2023-10-19
3.1 release compatibility 

mdupras:
### Prerequisites

### Description
I need a fix in the 3.1 releases, sadly the alpha version of the `.Web` isn't compatible with the 3.1 release yet. I created this branch to fix the `using` in the `.Web` package.

- Added Feedz.io source for alpha builds
- Updated some using to work with 3.1 releases

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -1,7 +1,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<configuration>
-  <packageSources>
-    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="feedz.io" value="https://f.feedz.io/sixlabors/sixlabors/nuget/index.json" />
-  </packageSources>
-</configuration>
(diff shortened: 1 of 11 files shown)


NuGet.Config
```diff
@@ -0,0 +1,7 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
JimBobSquarePants, line 1: This shouldn’t be required. The feed details are inherited in a props file via the shared infrastructure sub module

JimBobSquarePants:
Thanks for this. There’s build issues due to missing namespaces. Could you build this locally?

NuGet.Config
mdupras: Ohh good to know, removed it

mdupras:
> Thanks for this. There’s build issues due to missing namespaces. Could you build this locally?

Really weird, pretty sure it was building locally at some point. I justed fixed it.

image
I also have those changes, but I suspect its because I'm building on Linux, so I'm not including them in the PR.

JimBobSquarePants (approved):
LGTM 👍Thanks!

PR #352, opened 2024-01-15
Adding docker support

Lotto:
Adding Docker Support
Creating a Docker project to retrieve images from the /app/wwwroot folder using the ImageSharp service


JimBobSquarePants:
Hi @Lotto 

I’m not sure what purpose this PR serves as there’s been no [discussion](https://github.com/SixLabors/ImageSharp.Web/discussions). It looks like you’ve made unrelated changes also to the readme. 

Could you please open a discussion where you explain the use case. Thanks!

JimBobSquarePants:
Ah OK. I don’t think this belongs in this repository then. There’s too many configuration options for the middleware to be useful in enough cases. 

Thanks for your efforts though!!

sebastienros:
@Lotto I think you have a great idea and it deserves its own repository, while ImageSharp keeps being a non-opinionated framework only. You may be clear about the license implications though since the users of you image would be subject to it. But in the end it's good for everyone, people who want to use the power of ImageSharp, and for ImageSharp by providing ready-to-use solutions based on your knowledge.

Lotto:
Okay, so if I understood correctly, I'll close the pull request and keep the fork on my account that I will maintain.

Thank you!

PR #385, opened 2025-06-04
Updated AWSSDK.S3 to 4.0

mdupras:
### Prerequisites

### Description
Updated the AWSSDK.S3 to 4.0 release and fix the compilation issue.
Now `metadata.LastModified` can be null, I don't think it's like to happen in this scenario, but if it does the default Date is Now to assume the Image needs to be refreshed.

### DIFF
--- a/src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
+++ b/src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
@@ -9,6 +9,8 @@ namespace SixLabors.ImageSharp.Web;
 
 internal static class AmazonS3ClientFactory
 {
+    static AmazonS3ClientFactory() => AWSConfigs.InitializeCollections = true;
+
     /// <summary>
     /// Creates a new bucket under the specified account if a bucket
     /// with the same name does not already exist.


src/ImageSharp.Web.Providers.AWS/AmazonS3ClientFactory.cs
```diff
@@ -9,6 +9,8 @@ namespace SixLabors.ImageSharp.Web;
 
 internal static class AmazonS3ClientFactory
 {
+    static AmazonS3ClientFactory() => AWSConfigs.InitializeCollections = true;
```
JimBobSquarePants, line 12: Altering global behavior for transitive dependency does not appear be a prudent. It would be more appropriate to update the code to accommodate changes in behavior.

JimBobSquarePants:
@mdupras to follow up I discovered that there was a behavioral change in AWS. It's inserting a byte sequence before our output.

```
51 = '3'
50 = '2'
67 = 'C'
13 = '\r'
10 = '\n'
```

`32C\r\n` is `812` followed by line breaks which is the correct length of the blob. 

I added the following locally and everything works however I cannot get it to work in the CI on any OS. 

```
PutObjectRequest request = new()
{
    BucketName = this.bucketName,
    Key = this.GetKeyWithFolder(key),
    ContentType = metadata.ContentType,
    InputStream = stream,
    AutoCloseStream = false,
    UseChunkEncoding = false // THIS WAS DEFAULTING TO TRUE
};
```

PR #387, opened 2025-07-11
Update shared infra and fix styles

stefannikolei:


stefannikolei:
> We'll need to fix the build on main first (by copying some build config from ImageSharp)

And this project was still on net6

stefannikolei:
I have updated this lib to .net8 and .net9. I had to fix some issues which were reported.

The configure method was complaining multiple things because of aot. Because this is just a lil bit of syntactic sugar i went with deleting it

JimBobSquarePants:
@stefannikolei I need to have a good look at this as I'm not happy with all the new required attributes. Will try my best to get on it soon.

stefannikolei:
> @stefannikolei I need to have a good look at this as I'm not happy with all the new required attributes. Will try my best to get on it soon.

I was not happy about that dynamicaccessedattributes either. 

Those are probably from enabled aot. 

We could also go the way of removing them and add istrimmable = false.

stefannikolei:
I updated this branch. 

It's still to be discussed wether to disable trimming support or keep the added attributes

